PR #2, opened 2021-04-12
Switch to main thread before calling IMenuCommandService.AddCommand

### LINKED ISSUE #1: BaseCommand<T>.InitializeAsync should switch to main thread before calling commandService.AddCommand
The call to IMenuCommandService.AddCommand requires the main/UI thread so if the client is not already on the main thread when calling BaseCommand<T>.InitializeAsync then AddCommand will hang.

bluetarpmedia:
Fixes #1


madskristensen:
Thanks

PR #10, opened 2021-04-19
Build a project asynchronously

StevenRasmussen:
This PR adds a new extension method that allows building a project asynchronously and returns the build result as a `bool`.  To test this I have also added a new main menu and group that can be used for testing this and a good place for future commands for testing: 

image

If this is deemed useful I can also look at doing something similar for the solution.


madskristensen:
Awesome. Thank you so much!!

madskristensen:
I wonder if the old synchronous `Build` method should be removed now. I like to just have async methods when it makes sense, and I think it does here.

StevenRasmussen:
Agreed.  I didn't want to assume that was everyone's opinion so I left it in there.

PR #11, opened 2021-04-19
Added the ability to build the solution asynchronously

StevenRasmussen:
Added a new class for holding `EnvDTE.Solution` related extensions and added a method that allows building the solution asynchronously.  I also added a method to the `Community.VisualStudio.Toolkit.Solution` class that simply calls the extension method.

Added a new menu item to the test extension for testing:
image


madskristensen:
Thanks

PR #14, opened 2021-04-21
Verify that the package has a ProvideToolWindow attribute

bluetarpmedia:
While playing around with an extension and the toolkit I noticed that my command didn't open my tool window, and I soon found it was because I forgot the `ProvideToolWindow` attribute on the package. But new extenders may not realize this so quickly so I thought it would be useful to provide some feedback rather than silently failing.

This pull request modifies `BaseToolWindow<T>.Initialize` so that it throws if the package doesn't have a corresponding `ProvideToolWindow` attribute set for the tool window's Pane.

I also added a Debug.Fail but that will only fire for developers testing the toolkit itself -- not for users of the release-built nuget package.


madskristensen:
Fantastic idea! Thank you

PR #26, opened 2021-04-29
Created build target to check for presence of the Extensibility Essentials extension

reduckted:
From this discussion: https://github.com/VsixCommunity/Community.VisualStudio.Toolkit/discussions/13

This is the build target part. I'll leave this PR as a draft until the extension is written, which I aim to do this weekend. 🥳 

### Notes
* There are four different message levels that can be specified in the `ExtensibilityEssentialsLevel` build property:
   * `None` disables it.
   * `Info` (the default) produces a message.
   * `Warning` produces a build warning.
   * `Error` produces a build error.
* Any other value will produce a build warning telling you that the value is invalid.
* In Visual Studio 2019 (16.8+), the clickable link in the error list for _warnings and errors_ will take you to the marketplace page for the extension.
* For messages (the `Info` level), it's not possible to specify a link at this stage (see https://github.com/dotnet/msbuild/issues/6383)
* For Visual Studio 2017, a link cannot be provided, because that ability was only added in MSBuild 16.8 😢 (I assume it's not possible to make VS 2017 use a newer version of MSBuild).
* The message won't appear when building from MSBuild on the command line, or when building inside Visual Studio in a Continuous Integration environment (the message is suppressed when the `CI` environment variable is present).
* The message text can be overridden if you want to customize the message.

extensibility-essentials-message

### DIFF
--- a/Community.VisualStudio.Toolkit.sln
+++ b/Community.VisualStudio.Toolkit.sln
@@ -15,6 +15,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		src\AssemblyInfo.cs = src\AssemblyInfo.cs
 		src\Common.props = src\Common.props
 		src\Community.VisualStudio.Toolkit.props = src\Community.VisualStudio.Toolkit.props
+		src\Community.VisualStudio.Toolkit.targets = src\Community.VisualStudio.Toolkit.targets
 		README.md = README.md
 	EndProjectSection
 EndProject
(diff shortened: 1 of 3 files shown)


yannduran:
> I'll leave this PR as a draft until the extension is written

Umm,  what extension is that? Surely you're not writing an extension that has to be installed to point out that the _Extensibility Essentials_ extension isn't installed? :-P

I must be missing something LOL...

madskristensen:
This is absolutely awesome!!! I think the extension this belongs best in is the **Extensibility Essentials** extension pack itself.

Which reminds me, the Extensibility Essentials repo should be moved to the VSIX Community org

yannduran:
Gee thanks for clearing that up, guys. Clear as mud!

reduckted:
> Umm, what extension is that? Surely you're not writing an extension that has to be installed to point out that the Extensibility Essentials extension isn't installed? :-P

As I explained in the discussion #13:

> The Extensibility Essentials extension uses the IVsBuildPropertyStorage service to add a build property to every project when a solution is opened.

reduckted:
With the Extensibility Essentials extension change merged, I'd say this is good to go now.

madskristensen:
This is awesome. Thank you so much

PR #29, opened 2021-05-03
Add capability to BaseOptionModel to alter the default collection nam…

StevenBonePgh:
…e per-property via the use of an attribute. Add capability to BaseOptionModel to alter the default storage mechanism per property of 'serialize to string' to instead use the built-in SettingsStore supported types (string, bool, int, uint, long, ulong) via the use of an attribute. Add more in-depth documentation and logging to BaseOptionModel. Add overloads to ExceptionExtensions for logging messages or formatted strings along with the exception.

These changes are helpful for porting existing extensions to the toolkit where settings were stored using their underlying type and not serialized to base64 strings.


madskristensen:
Very clever. Thanks!

PR #30, opened 2021-05-03
Added extension methods for saving the solution.

StevenRasmussen:
These are certainly not ground breaking... the `SaveAsync` at least simplifies it for newcomers (don't need to pass in the name of the solution).  Really this is more about setting forth what the "ideal" would be which is to have an async version of being able to save the solution.  This at least brings that to the forefront as well as ensures that for now it's being handled properly by executing on the UI thread.


madskristensen:
Awesome. Thank you!

PR #31, opened 2021-05-04
ToolkitThreadHelper - a safe JoinableTaskFactory for MEF components (and other cases where package JTF not available)

bluetarpmedia:
Added:

- ToolkitThreadHelper.cs in Toolkit.Shared
- TextviewCreationListener.cs in the TestExtension 

`ToolkitThreadHelper` provides a `JoinableTaskFactory` that is safe to use in MEF components or when an `AsyncPackage.JoinableTaskFactory` is not available.

Usage example:

```
[Export(typeof(IWpfTextViewCreationListener))]
[ContentType("text")]
[TextViewRole(PredefinedTextViewRoles.PrimaryDocument)]
internal class TextviewCreationListener : IWpfTextViewCreationListener, IDisposable
{
    private readonly ToolkitThreadHelper _threadHelper = new ToolkitThreadHelper(ThreadHelper.JoinableTaskContext);

    public void Dispose()
    {
        Dispose(true);
        GC.SuppressFinalize(this);
    }

    protected virtual void Dispose(bool disposing)
    {
        if (disposing)
        {
            _threadHelper.Dispose();
        }
    }

    public void TextViewCreated(IWpfTextView textView)
    {
        // Note: RunAsync is not awaited/joined
        _threadHelper.JoinableTaskFactory.RunAsync(async () =>
        {
            try
            {
                await _threadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(_threadHelper.DisposalToken);

                // Do work, e.g.
                // await MyMethodAsync(_threadHelper.DisposalToken);
            }
            catch (Exception ex)
            {
                await ex.LogAsync();
            }
        });
    }
}
```

`JoinableTaskFactory` (hereafter 'JTF') from `ThreadHelper` is not the same as from an `AsyncPackage`. If you have an AsyncPackage available then always prefer to use its JTF instance over ThreadHelper. AsyncPackage's JTF will track all unjoined tasks and ensure they get joined when the package disposes before VS shuts down the CLR.
 
 In addition, AsyncPackage provides a disposal token so if you pass that token to all your async work (which you should) then VS can signal the token to cancel all work and ensure shutdown happens quickly, instead of having to wait for unfinished tasks to complete.

 But in some places (e.g. MEF components) you won't have an AsyncPackage available and may be tempted to use ThreadHelper's JTF. That's fine for Run and SwitchToMainThreadAsync and also for RunAsync as long as you await/join all tasks that RunAsync returns (although you still won't have a disposal token). But if you start fire-and-forget style tasks using `ThreadHelper.JoinableTaskFactory.RunAsync` and never explicitly await/join them then these tasks will never be joined.
 
 ToolkitThreadHelper solves the above by creating a new JTF instance along with a collection and a disposal token, and ensures that all unfinished tasks are joined during disposal, just like AsyncPackage.
 
 The implementation is based on aarnott's example here:
 https://github.com/microsoft/vs-threading/blob/main/doc/cookbook_vs.md#void-returning-fire-and-forget-methods

### DIFF
--- a/src/Community.VisualStudio.Toolkit.Shared/ToolkitThreadHelper.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/ToolkitThreadHelper.cs
@@ -1,0 +1,145 @@
+﻿using System;
+using System.Threading;
+using Microsoft.VisualStudio.Threading;
+
+namespace Community.VisualStudio.Toolkit.Shared
+{
+    /// <summary>
+    /// Provides a JoinableTaskFactory that is safe to use in MEF components or when an AsyncPackage.JoinableTaskFactory
+    /// is not available.
+    /// </summary>
+    /// <remarks>
+    /// JoinableTaskFactory (hereafter 'JTF') from ThreadHelper is not the same as from an AsyncPackage. If you have an
+    /// AsyncPackage available then always prefer to use its JTF instance over ThreadHelper. AsyncPackage's JTF will
+    /// track all unjoined tasks and ensure they get joined when the package disposes before VS shuts down the CLR.
+    /// 
+    /// In addition, AsyncPackage provides a disposal token so if you pass that token to all your async work (which you
+    /// should) then VS can signal the token to cancel all work and ensure shutdown happens quickly, instead of having
+    /// to wait for unfinished tasks to complete.
+    ///
+    /// But in some places (e.g. MEF components) you won't have an AsyncPackage available and may be tempted to use
+    /// ThreadHelper's JTF. That's fine for Run and SwitchToMainThreadAsync and also for RunAsync as long as you
+    /// await/join all tasks that RunAsync returns (although you still won't have a disposal token). But if you start
+    /// fire-and-forget style tasks using ThreadHelper.JoinableTaskFactory.RunAsync and never explicitly await/join them
+    /// then these tasks will never be joined.
+    /// 
+    /// ToolkitThreadHelper solves the above by creating a new JTF instance along with a collection and a disposal token,
+    /// and ensures that all unfinished tasks are joined during disposal, just like AsyncPackage.
+    /// 
+    /// The implementation is based on aarnott's example here:
+    /// https://github.com/microsoft/vs-threading/blob/main/doc/cookbook_vs.md#void-returning-fire-and-forget-methods
+    /// </remarks>
+    /// <example>
+    /// <code>
+    /// [Export(typeof(IWpfTextViewCreationListener))]
+    /// class TextviewCreationListener : IWpfTextViewCreationListener, IDisposable
+    /// {
+    ///     private readonly ToolkitThreadHelper _threadHelper = new(ThreadHelper.JoinableTaskContext);
+    ///     
+    ///     public void Dispose()
+    ///     {
+    ///         Dispose(true);
+    ///         GC.SuppressFinalize(this);
+    ///     }
+    ///     
+    ///     protected virtual void Dispose(bool disposing)
+    ///     {
+    ///         if (disposing)
+    ///         {
+    ///             _threadHelper.Dispose();
+    ///         }   
+    ///     }
+    /// 
+    ///     public void TextViewCreated(IWpfTextView textView)
+    ///     {
+    ///         // Note: RunAsync not awaited
+    ///         _threadHelper.JoinableTaskFactory.RunAsync(async () =>
+    ///         {
+    ///             try
+    ///             {
+    ///                 await _threadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(_threadHelper.DisposalToken);
+    ///                 
+    ///                 await MyMethodAsync(_threadHelper.DisposalToken);
+    ///             }
+    ///             catch (Exception ex)
+    ///             {
+    ///                 await ex.LogAsync();
+    ///             }
+    ///         });
+    ///     }
+    /// }
+    /// </code>
+    /// </example>
+    public class ToolkitThreadHelper : IDisposable
+    {
+        private readonly CancellationTokenSource _disposeCancellationTokenSource = new();
+        private readonly JoinableTaskContext _context;
+        private readonly JoinableTaskCollection _collection;
+        private bool _disposed = false;
+
+        /// <summary>
+        /// Creates the ToolkitThreadHelper.
+        /// </summary>
+        /// <param name="joinableTaskContext">
+        /// The application's one-and-only <see cref="JoinableTaskContext"/>.
+        /// For VS extensions, pass ThreadHelper.JoinableTaskContext, which is Visual Studio's singleton instance.
+        /// For unit tests, create a JoinableTaskContext instance (only one instance for the process) to supply here.
+        /// </param>
+        public ToolkitThreadHelper(JoinableTaskContext joinableTaskContext)
+        {
+            _context = joinableTaskContext;
+            _collection = joinableTaskContext.CreateCollection();
+            JoinableTaskFactory = joinableTaskContext.CreateFactory(_collection);
+        }
+
+        /// <summary>
+        /// The JoinableTaskFactory instance.
+        /// </summary>
+        public JoinableTaskFactory JoinableTaskFactory { get; }
+
+        /// <summary>
+        /// Gets a <see cref="CancellationToken"/> that can be used to check if the class has been disposed.
+        /// </summary>
+        public CancellationToken DisposalToken => _disposeCancellationTokenSource.Token;
+
+        /// <summary>
+        /// Joins any unfinished tasks.
+        /// </summary>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Joins any unfinished tasks.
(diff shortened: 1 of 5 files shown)


madskristensen:
Fantastic. Thank you so much!!

/cc @aarnott

src/Community.VisualStudio.Toolkit.Shared/ToolkitThreadHelper.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        /// <summary>
+        /// Creates the ToolkitThreadHelper.
+        /// </summary>
+        /// <param name="joinableTaskContext">
+        /// The application's one-and-only <see cref="JoinableTaskContext"/>.
+        /// For VS extensions, pass ThreadHelper.JoinableTaskContext, which is Visual Studio's singleton instance.
+        /// For unit tests, create a JoinableTaskContext instance (only one instance for the process) to supply here.
+        /// </param>
+        public ToolkitThreadHelper(JoinableTaskContext joinableTaskContext)
```
AArnott, line 88: We could probably do a default constructor that assumes `ThreadHelper.JoinableTaskContext`.
In fact even in unit tests this works when using https://aka.ms/vssdktestfx.

test/VSSDK.TestExtension/MEF/TextviewCreationListener.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Text.Editor;
+using Microsoft.VisualStudio.Utilities;
+
+namespace TestExtension.MEF
+{
+    [Export(typeof(IWpfTextViewCreationListener))]
+    [ContentType("text")]
+    [TextViewRole(PredefinedTextViewRoles.PrimaryDocument)]
+    internal class TextviewCreationListener : IWpfTextViewCreationListener, IDisposable
```
AArnott, line 13: nit: shouldn't there be a capital V here?
```diff
- internal class TextviewCreationListener : IWpfTextViewCreationListener, IDisposable
+ internal class TextViewCreationListener : IWpfTextViewCreationListener, IDisposable
```

test/VSSDK.TestExtension/MEF/TextviewCreationListener.cs
```diff
@@ -0,0 +1,50 @@
...
+                {
+                    await _threadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(_threadHelper.DisposalToken);
+
+                    // Do work, e.g.
+                    // await MyMethodAsync(_threadHelper.DisposalToken);
+                }
+                catch (Exception ex)
+                {
+                    await ex.LogAsync();
+                }
```
AArnott, line 46: Doesn't the toolkit already have a `LogAndForget` extension method that can be used on the result of `RunAsync` so it doesn't have to be done here, if this is all that the catch block will contain?

test/VSSDK.TestExtension/source.extension.vsixmanifest
```diff
@@ -1,20 +1,21 @@
...
-  <Dependencies>
-    <Dependency Id="Microsoft.Framework.NDP" DisplayName="Microsoft .NET Framework" d:Source="Manual" Version="[4.5,)" />
-  </Dependencies>
-  <Prerequisites>
-    <Prerequisite Id="Microsoft.VisualStudio.Component.CoreEditor" Version="[16.0,17.0)" DisplayName="Visual Studio core editor" />
-  </Prerequisites>
-  <Assets>
-    <Asset Type="Microsoft.VisualStudio.VsPackage" d:Source="Project" d:ProjectName="%CurrentProject%" Path="|%CurrentProject%;PkgdefProjectOutputGroup|" />
-  </Assets>
+    <Metadata>
```
AArnott, line 3: @madskristensen you should probably add an .editorconfig file to state and help the editor maintain the indentation size of this file.

AArnott (commented):
Very well done. Thank you for contributing, @bluetarpmedia.

src/Community.VisualStudio.Toolkit.Shared/ToolkitThreadHelper.cs
bluetarpmedia: Yeah, good idea! I tried it locally but I get a VSTHRD012 warning from the caller, because there are two overloads and it wants to prefer the one that takes the JoinableTaskContext. Any ideas? I think it's a good idea to keep the JoinableTaskContext parameter because it provides flexibility for usage outside of having access to the ThreadHelper or the SDK Test Framework. Is it possible to disable that warning from the class's side of things?

test/VSSDK.TestExtension/MEF/TextviewCreationListener.cs
bluetarpmedia: It has this extension method but only implemented for Task, not JoinableTask:

```
public static void ForgetAndLogOnFailure(this System.Threading.Tasks.Task task)
{
    task.ContinueWith(delegate (System.Threading.Tasks.Task antecedent)
    {
        antecedent.Exception!.Log();

    }, CancellationToken.None, TaskContinuationOptions.OnlyOnFaulted, TaskScheduler.Default).Forget();
}
```

Would an equivalent extension method for JoinableTask just need to do the same with the `JoinableTask.Task`?

test/VSSDK.TestExtension/MEF/TextviewCreationListener.cs
bluetarpmedia: Thanks, have sent a new PR.

src/Community.VisualStudio.Toolkit.Shared/ToolkitThreadHelper.cs
bluetarpmedia: Two static factory methods with a private ctor would do it:

```
public static ToolkitThreadHelper Create()
{
    return new ToolkitThreadHelper(ThreadHelper.JoinableTaskContext);
}

public static ToolkitThreadHelper CreateWithContext(JoinableTaskContext joinableTaskContext)
{
    return new ToolkitThreadHelper(joinableTaskContext);
}
```

AArnott: There's no way to suppress the analyzer from the ctor overloads unless you add `[Obsolete]` to the one that takes JTC, and you don't want to do that.
Your approach with the static factory methods is interesting.

test/VSSDK.TestExtension/MEF/TextviewCreationListener.cs
AArnott: Yes, that would do it. You could also do it without an additional extension method like this:

```cs
).Task.ForgetAndLogOnFailure();
```

src/Community.VisualStudio.Toolkit.Shared/ToolkitThreadHelper.cs
bluetarpmedia: I've updated the PR with the static factory methods and new ForgetAndLogOnFailure overload for JoinableTask which the TextViewCreationListener example uses.

yannduran:
The name `ForgetAndLogOnFailure` is both backwards and a bit verbose.

Wouldn't just `LogAndForget` make more sense, as you're first going to **_log_** the event, then **_forget_** anything after that? And it'd be more succinct without losing meaning?

reduckted:
> Wouldn't just LogAndForget make more sense, as you're first going to log the event, then forget anything after that? And it'd be more succinct without losing meaning?

I disagree. The first thing that you are doing is forgetting about the task, not logging the task, so `Forget` being first makes sense. The logging only happens on failure, so the `LogOnFailure` part also makes sense.

It's a bit difficult to indicate what actually happens without any punctuation. Something like "Forget about this task. Also log if the task fails" would be more descriptive, but alas, that's not a valid member name. 😆 

I mentioned this in another comment, but `static void Forget(this Task task, bool logOnFailure = true);` would be good. You do lose the "and log if there's an error" part of the name, but at least it's clear that you are forgetting about the task as one step, and as a separate step, you can have it log something if there's a failure.

StevenBonePgh:
> static void Forget(this Task task, bool logOnFailure = true); would be good.
I'll disagree with this suggestion. The `ForgetAndLogOnFailure`, while verbose, is accurate and descriptive, and with IntelliSense the verbosity is a non-issue when weighed against descriptive. The `ForgetAndLogOnFailure` extension method would allow for other overloads, such as taking a `string message` and `string format, params object?[] args` parameters. I recently added these overloads to `ExceptionExtensions`, as sometimes (most times even) diagnosing an Exception without a bit more context is near impossible.

yannduran:
Well, while I understand where you're both coming from, we'll have to agree to disagree. While it might be technically correct, it just sounds clumsy and **_feels_** wrong. I just don't like it, sorry.

Don't forget, newcomers are going to come across `FileAndForget` in documentation and/or videos which also logs on failure, but doesn't say so in the name. I'm no newcomer and the name didn't make any sense to me until it was spelled out.

There's _technically correct_, and then there's _**easy to remember/use**_. That's kind of the whole point of the toolkit in the first place, isn't it?

PR #32, opened 2021-05-04
Changes to ToolkitThreadHelper & TextViewCreationListener and new ForgetAndLogOnFailure overload for JoinableTask

bluetarpmedia:
Changes after code review:
* New ForgetAndLogOnFailure(JoinableTask) overload
* Replaced ToolkitThreadHelper ctor with two static factory methods
* TextViewCreationListener (example of how to use new ToolkitThreadHelper) now uses ForgetAndLogOnFailure
* Fixed TextViewCreationListener case

### DIFF
--- a/src/Community.VisualStudio.Toolkit.Shared/ExtensionMethods/TaskExtensions.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/ExtensionMethods/TaskExtensions.cs
@@ -5,13 +5,17 @@ using Microsoft.VisualStudio.Threading;
 
 namespace Microsoft.VisualStudio.Shell
 {
-    /// <summary>Extension methods for the <see cref="System.Threading.Tasks.Task" /> object.</summary>
+    /// <summary>
+    /// Extension methods for <see cref="System.Threading.Tasks.Task" /> and <see cref="JoinableTask" />.
+    /// </summary>
     public static class TaskExtensions
     {
         /// <summary>
-        /// Starts a Task and lets it run in the background, while silently handles any exceptions.
+        /// Logs error information to the Output Window if the given <see cref="System.Threading.Tasks.Task" /> faults.
         /// </summary>
         /// <remarks>
+        /// The task itself starts before this extension method is called and only continues in
+        /// this extension method if the original task throws an unhandled exception.
         /// This is similar to the <c>task.Forget()</c> method, but it logs any unhandled exception
         /// thrown from the task to the Output Window.
         /// </remarks>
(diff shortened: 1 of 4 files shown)


AArnott (approved):

yannduran:
What happened to `static void Forget(this Task task, bool logOnFailure = true)`? Andrew gave it a thumbs up, but the latest toolkit version seems to still have `ForgetAndLogOnFailure`?

yannduran:
Btw, I noticed that in my earlier suggestion that it should have said `LogOnFailureAndForget` not `LogAndForgetOnFailure`, because as @reduckted mentioned `LogAndForgetOnFailure` would have been misleading.

madskristensen:
It can't be called `Forget()` since that method already exist in one of the threading libraries.

StevenBonePgh:
> What about just static void Forget(this Task task, bool logOnFailure = true)?

Copied from other thread: I'll disagree with this suggestion. Adding the `logOnFailure` parameter to the `Forget()` extension method precludes allowing for other discoverably named overloads that log on failure, such as taking a `string message` and `string format, params object?[] args` parameters. I recently added these overloads to `ExceptionExtensions`, as sometimes (most times even) diagnosing an Exception without a bit more context is near impossible.

The current `ForgetAndLogOnFailure` has the bonus of showing up in IntelliSense in the right order when you start typing Forget vs the `LogOnFailureAndForget` suggestion. First and foremost you are 'forgetting' and then, on failure, it is logged.

AArnott:
> It can't be called Forget() since that method already exist in one of the threading libraries.

Why does that forbid it? I thought multiple extension method overloads could come from different libraries.

yannduran:
@StevenBonePgh 

>Adding the logOnFailure parameter to the Forget() extension method precludes allowing for other discoverably named overloads that log on failure, such as taking a string message and string format, params object?[] args parameters

I don't think that **_possible_** other overloads should be the determining factor for overly complicating the name of the method. 

> I recently added these overloads to ExceptionExtensions

If you're adding extension methods, then maybe **they** should be the ones with a more complicated name.

>The current ForgetAndLogOnFailure has the bonus of showing up in IntelliSense in the right order

Your argument about discoverability is not really valid. Typing `.forget`shows all methods with 'forget' **in** them.

image

>First and foremost you are 'forgetting' and then, on failure, it is logged.

You're arguing **_technicalities_** whereas the toolkit was supposed to make extensions **easier** to write. The logging should be handled behind the scenes without distracting developers who are new to writing extensions. And for more advanced scenarios the optional parameter would allow turning the logging off.

I don't mind whether it's called `Forget()` or `FireAndForget()`, but I **strenuously** disagree with it being called something as verbose as `ForgetAndLogOnFailure`. 

I don't want to see this toolkit devolve into having **complicated** names (for the sake of being more **_technically_** correct). The original intent of the toolkit was to keep things **simple**.

PR #37, opened 2021-05-14
OutputWindowPane

bluetarpmedia:
`OutputWindowPane` allows an extension to create a new Output window pane or get an existing one.
A pane can be activated (shown), hidden and cleared. Text can be written to the pane via methods like `WriteLineAsync` or with a `TextWriter` returned from `CreateOutputPaneTextWriterAsync`. (For newer versions of VS, this uses [OutputWindowTextWriter](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.outputwindowtextwriter?view=visualstudiosdk-2019) from the SDK.)

Examples:
```
OutputWindowPane generalPane = await VS.Windows.GetOutputWindowPaneAsync(VSOutputWindowPane.General);
OutputWindowPane buildPane = await VS.Windows.GetOutputWindowPaneAsync(VSOutputWindowPane.Build);
OutputWindowPane debugPane = await VS.Windows.GetOutputWindowPaneAsync(VSOutputWindowPane.Debug);

Guid myPaneGuid;
{
    OutputWindowPane pane = await VS.Windows.CreateOutputWindowPaneAsync("My Pane");
    myPaneGuid = pane.Guid;
    await pane.WriteLineAsync("My message");
}

// Elsewhere:
{
    OutputWindowPane pane = await VS.Windows.GetOutputWindowPaneAsync(myPaneGuid);
    pane.Name = "My Pane - Updated";

    using (TextWriter writer = await pane.CreateOutputPaneTextWriterAsync())
    {
        char[] buffer = GetSomeChars();
        await writer.WriteLineAsync("This is a more efficient way to write lots of text.");
        await writer.WriteAsync(buffer, 0, buffer.Length);
        await writer.WriteLineAsync();
    }
}
```

### DIFF
--- a/src/Community.VisualStudio.Toolkit.Shared/Helpers/OutputWindowPane.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/Helpers/OutputWindowPane.cs
@@ -1,0 +1,276 @@
+﻿using System;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Threading;
+using Task = System.Threading.Tasks.Task;
+
+namespace Community.VisualStudio.Toolkit.Shared.Helpers
+{
+    /// <summary>
+    /// A pane in the Output window.
+    /// </summary>
+    /// <remarks>
+    /// OutputWindowPane allows an extension to create a new Output window pane or get an existing one.
+    /// A pane can be activated (shown), hidden and cleared. Text can be written to the pane via methods
+    /// like <see cref="WriteLineAsync(string)"/> or with a <see cref="TextWriter"/> returned
+    /// from <see cref="CreateOutputPaneTextWriterAsync"/>.
+    /// </remarks>
+    /// <example>
+    /// <code>
+    /// Guid myPaneGuid;
+    /// {
+    ///     OutputWindowPane pane = await VS.Windows.CreateOutputWindowPaneAsync("My Pane");
+    ///     myPaneGuid = pane.Guid;
+    ///     await pane.WriteLineAsync("My message");
+    /// }
+    /// 
+    /// // Elsewhere:
+    /// {
+    ///     OutputWindowPane pane = await VS.Windows.GetOutputWindowPaneAsync(myPaneGuid);
+    ///     pane.Name = "My Pane - Updated";
+    ///     using (TextWriter writer = await pane.CreateOutputPaneTextWriterAsync())
+    ///     {
+    ///         char[] buffer = GetSomeChars();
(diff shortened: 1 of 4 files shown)


madskristensen:
That’s awesome. Thank you so much. We should update the *LogAsync* extension methods to use this new output window helper

bluetarpmedia:
> We should update the _LogAsync_ extension methods to use this new output window helper

Yes, great idea! I've submitted a new PR for that.

src/Community.VisualStudio.Toolkit.Shared/Helpers/OutputWindowPane.cs
```diff
@@ -0,0 +1,276 @@
...
+        /// <summary>
+        /// Uniquely identifies the Output window pane.
+        /// After creating a pane, you can cache this Guid and later obtain the same pane from <see cref="GetAsync(Guid)"/>.
+        /// </summary>
+        public Guid Guid { get; }
+
+        /// <summary>
+        /// The name (title) of the Output window pane.
+        /// </summary>
+        public string Name
```
madskristensen, line 114: What's the use case for changing the name of the output window pane? This seems like a bad UX practice

yannduran:
While I usually create a custom pane in my extensions, I have never needed/wanted to change the name later, so maybe that's a good point.

bluetarpmedia:
>What's the use case for changing the name of the output window pane? This seems like a bad UX practice

>While I usually create a custom pane in my extensions, I have never needed/wanted to change the name later, so maybe that's a good point.

Yeah, I've never needed to do so either, but I think there's an interesting design decision to make here.

Just because the functionality may provide a bad usability experience for the end-user (the person using the extension), should the Toolkit avoid providing APIs that allow an extension author to make that choice?

E.g. a similar argument could be made about allowing an extension to launch a ridiculous number of tool windows, or fill up menus with poorly named commands, or place a command named `Build Solution` under the `Help` menu.

Like I said, I've never needed to rename a Pane, but maybe there's some reason some author will want to do so. E.g. update its name based on some state in the project, or last successful build time. (Is that good UX? Maybe not, but I'd say it's that author's choice.)

I don't mean to suggest the Toolkit should provide poorly-designed APIs for the extension programmer to use. But I think the functionality that VS itself provides is different to whether the API is well-designed and clean/discoverable/documented/etc. In this case, a public property `Name` with getter/setter is as good as it gets, IMO.

Despite all the above, however, I don't mean to suggest I feel strongly about the Name setter. I personally don't think I'll ever need it and would be happy for it to go. I'm more interested in the larger design decision: if the API is designed well, and the functionality is available in VS, then should we be making decisions that prevent extension authors from using it?

yannduran:
If one of the goals of the toolkit is to encourage best practices, I'd say it's not a good idea to provide a property that nobody is likely to want to use (if they care about their UX) just for the sake having it there. 

If someone comes up with a use case I guess it could be added at a later stage?

PR #41, opened 2021-05-19
Added the ability to show a 'System.Windows.Window' as a dialog.

StevenRasmussen:
Per the [documentation](https://docs.microsoft.com/en-us/visualstudio/extensibility/creating-and-managing-modal-dialog-boxes?view=vs-2019), showing a dialog window properly takes a little more code than just `window.ShowDialog`.  This PR encapsulates that logic in an extension method and also a helper method on the `VS.Windows` class.

### DIFF
--- a/src/Community.VisualStudio.Toolkit.Shared/ExtensionMethods/WindowExtensions.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/ExtensionMethods/WindowExtensions.cs
@@ -1,0 +1,43 @@
+﻿using System;
+using System.Threading.Tasks;
+using System.Windows;
+using Microsoft;
+using Microsoft.Internal.VisualStudio.PlatformUI;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Community.VisualStudio.Toolkit.Shared.ExtensionMethods
+{
+    /// <summary>Extension methods for the <see cref="Window"/> class.</summary>
+    public static class WindowExtensions
+    {
+        /// <summary>Shows a window as a dialog.</summary>
+        public static async Task<bool?> ShowDialogAsync(this Window window, WindowStartupLocation windowStartupLocation = WindowStartupLocation.CenterOwner)
+        {
+            if (window == null)
+                throw new ArgumentNullException(nameof(window));
+
+            var vsUiShellService = await VS.GetServiceAsync<SVsUIShell, IVsUIShell>();
+            Assumes.Present(vsUiShellService);
+
+            if (vsUiShellService.GetDialogOwnerHwnd(out var hwnd) != VSConstants.S_OK)
+                throw new Exception("Unable to get the dialog owner handler.");
+
+            if (vsUiShellService.EnableModeless(0) != VSConstants.S_OK)
+                throw new Exception("Unable to set the UI Shell Service to 'Modeless'.");
+
+            window.WindowStartupLocation = windowStartupLocation;
+
+            try
+            {
+                WindowHelper.ShowModal(window, hwnd);
+                return window.DialogResult;
+            }
+            finally
+            {
+                vsUiShellService.EnableModeless(1);
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


ErikEJ:
I always just use DialogWindow ...

reduckted:
From the documentation:

> There are two ways of doing this. The recommended way, if you have a WPF dialog box, is to derive it from `DialogWindow`, and then call `ShowModal` to display the dialog box.

Do we really want to be making it easier to do things that are not recommended?

src/Community.VisualStudio.Toolkit.Shared/ExtensionMethods/WindowExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <summary>Shows a window as a dialog.</summary>
+        public static async Task<bool?> ShowDialogAsync(this Window window, WindowStartupLocation windowStartupLocation = WindowStartupLocation.CenterOwner)
+        {
+            if (window == null)
+                throw new ArgumentNullException(nameof(window));
+
+            var vsUiShellService = await VS.GetServiceAsync<SVsUIShell, IVsUIShell>();
+            Assumes.Present(vsUiShellService);
+
+            if (vsUiShellService.GetDialogOwnerHwnd(out var hwnd) != VSConstants.S_OK)
```
reduckted, line 24: Using `ErrorHandler.ThrowOnFailure` would be a better choice here (and below), as it will get the error message based on the error code.

```cs
ErrorHandler.ThrowOnFailure(vsUiShellService.GetDialogOwnerHwnd(out var hwnd));
```

yannduran:
I've been meaning to suggest replacing the `VSUtilities.ShowMessage()` method with a custom dialog, because it has limited `OLEMSGICON` options, and should really be something WPF based anyway. The results should also be a more familiar WPF-based value than the COM-based stuff we're trying to hide.

So this would help with that too.

Doesn't using `DialogWindow .ShowModal()` do the parenting of the window to VS behind the scenes? I mean, isn't that why it exists?

src/Community.VisualStudio.Toolkit.Shared/ExtensionMethods/WindowExtensions.cs
yannduran: That wouldn't be needed if it was based on WPF's `DialogWIndow` instead of the COM-based `VSUtilities.ShowMessage()`..

StevenRasmussen:
@reduckted - In some instances it may not be feasible to inherit from the `DialogWindow` class.  I use a 3rd party library from DevExpress which has it's own window class.  Using this window along with all of their controls automatically handles all the theming of controls... they have themes that correspond with the 3 major VS themes and so it is pretty seamless to make the UI close to native VS controls/theming.  So I think there are still use cases where we should provide this alternative.

src/Community.VisualStudio.Toolkit.Shared/ExtensionMethods/WindowExtensions.cs
StevenRasmussen: @reduckted - Thanks for pointing this out.  Wasn't aware of the helper method.  Updated now.

reduckted:
>  Using this window along with all of their controls automatically handles all the theming of controls

Not sure if it handles the theming of the DevExpress controls, but do you know about the `UseVSTheme` attached property?

https://github.com/VsixCommunity/Community.VisualStudio.Toolkit/blob/26bdbf6ee28d0c706c2ad92fe4f226b9ebbddee5/test/VSSDK.TestExtension/ToolWindows/ThemeWindowDialog.xaml#L15

StevenRasmussen:
I saw that there was some general theming support in the toolkit but I'm so heavily invested with DevExpress right now (44 forms and 29 custom user controls) it doesn't really make sense for me to try and use any of the native theming support.  I do handle the 'ThemeChanged' event to trigger all of the DevExpress controls to update based on the current theme. Thanks though.

madskristensen:
This sounds like a subject we need to document in the https://github.com/VsixCommunity/docs repo

StevenRasmussen:
@madskristensen - Is this PR deemed not useful then?  Or are you wanting a PR to the docs repo before accepting this?  It's not clear to me what the status is on this. Thanks

madskristensen:
In cool with this, but there are merge conflicts

StevenRasmussen:
Updated with the latest from master.  Should be good to merge at this point. Thanks

madskristensen:
Thanks

PR #42, opened 2021-05-20
Added new 'GetRequiredServiceAsync' method.

StevenRasmussen:
Based on the XML comments for the `VS.GetServiceAsync` method, it should return `null` if the service is not found... but that's not true. The `Assumes.Present` method will throw an exception if the service is not found. This PR removes that line and creates a new `VS.GetRequiredServiceAsync` method which does perform that logic and follows the current pattern of the `System.IServiceProvider` naming conventions.

### DIFF
--- a/src/Community.VisualStudio.Toolkit.Shared/VS.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/VS.cs
@@ -37,7 +37,7 @@ namespace Community.VisualStudio.Toolkit
         public static Windows Windows => new();
 
         /// <summary>Get the EnvDTE which provide a broad API for a large part of Visual Studio.</summary>
-        public static Task<DTE2> GetDTEAsync() => GetServiceAsync<DTE, DTE2>();
+        public static Task<DTE2> GetDTEAsync() => GetRequiredServiceAsync<DTE, DTE2>();
 
         /// <summary>
         /// Gets a global service asynchronously.
--- a/src/Community.VisualStudio.Toolkit.Shared/VS.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/VS.cs
@@ -45,7 +45,7 @@ namespace Community.VisualStudio.Toolkit
         /// <typeparam name="TService">The type identity of the service.</typeparam>
         /// <typeparam name="TInterface">The interface to cast the service to.</typeparam>
         /// <returns>A task who's result is the service, if found; otherwise <see langword="">null</see>.</returns>
-        public static async Task<TInterface> GetServiceAsync<TService, TInterface>() where TService : class where TInterface : class
+        public static async Task<TInterface?> GetServiceAsync<TService, TInterface>() where TService : class where TInterface : class
         {
 #if VS14
             return (TInterface)await AsyncServiceProvider.GlobalProvider.GetServiceAsync(typeof(TService));
--- a/src/Community.VisualStudio.Toolkit.Shared/VS.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/VS.cs
@@ -48,6 +48,7 @@ namespace Community.VisualStudio.Toolkit
         public static async Task<TInterface?> GetServiceAsync<TService, TInterface>() where TService : class where TInterface : class
         {
 #if VS14
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
             return (TInterface)await AsyncServiceProvider.GlobalProvider.GetServiceAsync(typeof(TService));
 #elif VS15
             return await ServiceProvider.GetGlobalServiceAsync<TService, TInterface>();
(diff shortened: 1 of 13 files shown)


madskristensen:
@aarnott we followed the current pattern on purpose IIRC. That's also why the method doesn't return a nullable value. Do you remember why we did that and if there's anything here to be aware of? The PR seems reasonable to me, but I'm afraid I'm missing something...

madskristensen:
What's the use case of the `GetRequiredServiceAsync` method. Is it for internal use by the toolkit itself?

StevenRasmussen:
My thought process behind this PR was initially sparked by the XML comments for the `GetServiceAsync` method stating that it would return `null` if the service was not found.  I saw that it called the `Assumes.Present` method which will throw if the service doesn't exist.  I've been working quite a bit with the `System.IServiceProvider` (used by default with .Net Core/.Net 5/6 Asp.Net projects for DI) which has the 'GetService` method which does not throw if the service is not found but has an extension method called `GetRequiredService` which does throw.  So I thought that in the world of .Net where the `System.IServiceProvider` is quite well known and used then we should try and follow that same paradigm here.

As for the use case: sometimes you may be searching for a service as a way to determine if a certain feature is supported or not and have branching logic based on the availability of that service... so the `GetServiceAsync` method would apply since you don't necessarily want to throw an exception in this case.  In other instances, you may be depending on a service to always be available and so you would call the `GetRequiredServiceAsync` method and want to throw if it's not available.

madskristensen:
That makes sense, and the required version of the method should then be used everywhere internally by the toolkit itself.

madskristensen:
Before taking this PR, I hope @aarnott will chime in on this

madskristensen:
@aarnott any concerns?

yannduran:
@madskristensen Would it be possible for us to be present at these Teams reviews that you mentioned? 

Mainly to just listen & learn, but maybe also to be able to answer any questions or offer opinions if asked. It would be so valuable to actually see/hear the process that you and Andrew go through!

madskristensen:
@yannduran at the end of our last review, I mentioned to Andrew that it was a shame we didn't record it. We'll do that next time

src/Community.VisualStudio.Toolkit.Shared/VS.cs
```diff
@@ -47,13 +48,25 @@ public static class VS
         public static async Task<TInterface> GetServiceAsync<TService, TInterface>() where TService : class where TInterface : class
```
AArnott, line 48: Now that this may return null, the nullable type ref annotation must be changed:

```suggestion
        public static async Task<TInterface?> GetServiceAsync<TService, TInterface>() where TService : class where TInterface : class
```

AArnott: Also the doc line above it uses incorrect syntax for the `null` keyword. It should be:

```diff
-/// <returns>A task who's result is the service, if found; otherwise <see langword="">null</see>.</returns>
+/// <returns>A task who's result is the service, if found; otherwise <see langword="null" />.</returns>

File: src/Community.VisualStudio.Toolkit.Shared/VS.cs [hunk 2]
```diff
@@ -47,13 +48,25 @@ public static class VS
...
+            return await ServiceProvider.GetGlobalServiceAsync<TService, TInterface>(swallowExceptions: false);
 #endif
+        }
+
+        /// <summary>
+        /// Gets a global service asynchronously.
+        /// </summary>
+        /// <typeparam name="TService">The type identity of the service.</typeparam>
+        /// <typeparam name="TInterface">The interface to cast the service to.</typeparam>
+        /// <returns>A task who's result is the service, if found; otherwise throws an exception.</returns>
```
AArnott, line 64:
```suggestion
        /// <returns>A task who's result is the service, if found.</returns>
        /// <exception cref="Exception">Throws an exception when the service is not available.</exception>
```

src/Community.VisualStudio.Toolkit.Shared/VS.cs
```diff
@@ -47,13 +48,25 @@ public static class VS
         public static async Task<TInterface> GetServiceAsync<TService, TInterface>() where TService : class where TInterface : class
         {
 #if VS14
-            var service = (TInterface)await AsyncServiceProvider.GlobalProvider.GetServiceAsync(typeof(TService));
+            return (TInterface)await AsyncServiceProvider.GlobalProvider.GetServiceAsync(typeof(TService));
```
AArnott, line 51: Unrelated to your change, but this is buggy because it can lead to the cast of an STA COM object on a background thread.
```suggestion
            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
            return (TInterface)await AsyncServiceProvider.GlobalProvider.GetServiceAsync(typeof(TService));
```

The other two sections are OK because the methods that take two generic args are responsible to do the thread switch.

AArnott (changes requested):
No concerns in general approach. Just a few concern with the impl of the changes and a couple of surrounding issues I noticed.

StevenRasmussen:
Just committed all the suggestions.  Anything else?

madskristensen:
Awesome. Thanks everybody

PR #43, opened 2021-05-22
Project type Guid fields

reduckted:
The list of [project type GUID strings](https://github.com/VsixCommunity/Community.VisualStudio.Toolkit/blob/e39138f93a32cb3f37e6f11ad766930acd8036f7/src/Community.VisualStudio.Toolkit.Shared/Helpers/ProjectTypes.cs) is awesome, but sometimes you need to compare a `Guid` with a project type. You currently have two options:
1. Convert the `Guid` to a string and do a string comparison, but the format of the GUID matters in that case, or
2. Parse the string constant to a `Guid`, but that's a bit ugly and there's a (very) small performance overhead of parsing the GUID.

Having a `Guid` field for each project type would be ideal (in addition to the string constants, of course).

I thought this would be a great opportunity to use a C# source generator that starts with the existing `ProjectTypes` class and creates a class that has `Guid` fields instead of `string` constants. If any new project types are added to the `ProjectTypes` class, the source generator will automatically create the `Guid` field, so we don't have to keep the two classes in sync manually.

The only downside to using a source generator is that it cannot be used in the v14 and v15 projects because they target `net46` and the source generator project needs to be `netstandard2.0`, which can only be referenced by `net461` projects and higher. What that means is the new class that contains the project type `Guid` fields will only be in the v16+ versions of the toolkit. Not a deal-breaker in my opinion.

I've nested the new class inside the existing `ProjectTypes` class because it felt better than having another class that had a similar name. To use an existing string constant you would do this:
```cs
string s = EnvDTE.ProjectTypes.ASPNET_CORE;
```
To use the new `Guid` field you would do this:
```cs
Guid g = EnvDTE.ProjectTypes.Guids.ASPNET_CORE;
```

Here's what _Class View_ shows for the existing `ProjectTypes` and the new `ProjectTypes.Guids` classes.
image
image


StevenRasmussen:
Great idea! I haven't worked with source generators yet but it's my understanding that the output is a real file somewhere (though I don't remember where goes... maybe in the bin or obj folders)... for VS14 and VS15 does it make sense to add the output as a linked file in those projects so that they still get the benefit?  Just thinking out loud here.

yannduran:
That's a good idea in my opinion. I use that technique in several projects in a solution to have the Vsix.Name etc constants available to them. I've had no problem doing that. Once it's set up it just works perfectly.

madskristensen:
Why not convert the strings in the `ProjectTypes` class to Guids instead of having two separate classes? Then overload the `Project.IsKind` extension method to take a Guid parameter instead of just strings.

reduckted:
> Why not convert the strings in the `ProjectTypes` class to Guids instead of having two separate classes?

Without constants, we couldn't [use the project types in attributes](https://github.com/VsixCommunity/ExtensibilityEssentials/blob/master/src/Shared/ExtensibilityEssentialsPackage.cs#L20).

Plus, if you did need to do a string comparison for whatever reason, you would need to know what format to use, and the default `Guid.ToString()` output is _not_ the same as the constant value.

```cs
EXTENSIBILITY;                     // {82B43B9B-A64C-4715-B499-D71E9CA2BD60}
Guids.EXTENSIBILITY.ToString();    // 82b43b9b-a64c-4715-b499-d71e9ca2bd60
Guids.EXTENSIBILITY.ToString("n"); // 82b43b9ba64c4715b499d71e9ca2bd60
Guids.EXTENSIBILITY.ToString("d"); // 82b43b9b-a64c-4715-b499-d71e9ca2bd60
Guids.EXTENSIBILITY.ToString("b"); // {82b43b9b-a64c-4715-b499-d71e9ca2bd60}
Guids.EXTENSIBILITY.ToString("p"); // (82b43b9b-a64c-4715-b499-d71e9ca2bd60)
Guids.EXTENSIBILITY.ToString("x"); // {0x82b43b9b,0xa64c,0x4715,{0xb4,0x99,0xd7,0x1e,0x9c,0xa2,0xbd,0x60}}
```

reduckted:
> I haven't worked with source generators yet but it's my understanding that the output is a real file somewhere

It's all in memory by default, and the documentation is a bit lacking, but I did manage to find [how to get the generated files written to disk](https://til.cazzulino.com/dotnet/persisting-output-files-from-source-generators)! 🎉 

> for VS14 and VS15 does it make sense to add the output as a linked file in those projects so that they still get the benefit? Just thinking out loud here.

It's very hacky, but it does seem to get the job done. 😁 We just need to make sure the v14 and v15 projects build _after_ v16 and v17, otherwise they may not use the latest generated code. I believe we can do that by configuring the project dependencies, which in turn controls the build order of projects. I've tried it and it appears to work. If we think that's a good solution, then I'll push up those changes.

yannduran:
I haven't had a chance to look into this yet, but why couldn't we just have both the **_guid_** and the **_string_** values like Mads' Vsix Synchronizer produces? I'm not suggesting they be generated at all, but simply have both values in the same class.

image

reduckted:
> I like @yannduran suggestion the best here. It's simple, easy to understand and maintain.

It could be argued that it's _less_ easy to maintain since adding a new GUID requires two lines of code instead of one. 😜 

> If so, we have the `Project.IsKind(string guid)` extension method for this very purpose. Can't you use that?

So the only time anyone would ever need to use a project type GUID is when they're using a `Project` object? That does seem a bit limiting, but if you think that's the only place you'd use a project type GUID, then it would would probably be sufficient.

PR #59, opened 2021-06-10
Fixed foreground and background colors for theming

### LINKED ISSUE #57: Themes.UseVsTheme doesn't appear to apply to TextBlock
I'm using the Blue theme in Visual Studio. If I add a TextBlock containing text to ThemeWindowDemo.xaml as follows
```
            <TextBlock Margin="{StaticResource Margin}">
                <TextBlock Text="Click here: " />
                <Hyperlink>Interesting link</Hyperlink>
            </TextBlock>
```

The TextBlock is barely legible. I suspect it is rendering as almost-white text on white background (sample image below). Would it be appropriate for TextBlocks to be handled by Themes.UseVsTheme?

image

reduckted:
Fixes #57.

`EnvironmentColors.StartPageTextBodyBrushKey` looked OK when using the dark theme, but it was basically white when using the light and blue themes, so it's not a suitable foreground color.

I've change the foreground color to use `ThemedDialogColors.WindowPanelTextBrushKey`, which appears to be what the `Label` uses:

image
(light, blue and dark themes from top to bottom).

To match this change, I've also changed the background color to `ThemedDialogColors.WindowPanelBrushKey`. This _does_ make the background slightly darker for the light and blue themes. The dark theme appears to be the same color.

Before:
image
After:
image

### DIFF
--- a/src/Community.VisualStudio.Toolkit.Shared/Themes/Themes.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/Themes/Themes.cs
@@ -49,8 +49,8 @@ namespace Community.VisualStudio.Toolkit
                 {
                     if ((bool)e.NewValue)
                     {
-                        OverrideProperty(element, Control.BackgroundProperty, _originalBackgroundProperty, EnvironmentColors.StartPageTabBackgroundBrushKey);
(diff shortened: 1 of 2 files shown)


madskristensen:
Awesome. Thanks

PR #62, opened 2021-06-18
Change tool window nested Pane classes to internal instead of public

bluetarpmedia:
When FxCop/NetAnalyzers are enabled, the nested `public class Pane` causes warning `CA1034: Nested types should not be visible`

This change is mainly for the benefit of people copying these samples, since these analyzers are not enabled in the Toolkit itself.


reduckted:
Would be great to have the same change made in the template pack. 👍

madskristensen:
Thanks

bluetarpmedia:
> Would be great to have the same change made in the template pack. 👍

Yeah for sure! I've submitted a PR for this: https://github.com/VsixCommunity/ExtensibilityTemplatePack/pull/9

PR #95, opened 2021-07-04
KnownCommands

### LINKED ISSUE #93: Predefined CommandID values for known commands
The new `Commands` class that lets you execute a command by name and the overloads for `VSConstants.VSStd97CmdID` and so on are fantastic, but knowing which enum the command you want to execute is defined in makes it a bit painful to use. 

I think a `KnownCommands` class that defines the `CommandID` objects for all commands that Visual Studio defines would be great. I've created a proof-of-concept here: https://github.com/VsixCommunity/Community.VisualStudio.Toolkit/compare/master...reduckted:feature/known-commands

The good thing is that the `CommandID`s for all 800+ commands that Visual Studio provides can be generated quite easily (see the PowerShell script).

Here's what you see in intellisense:
image

There's still a few things to improve, but I thought I'd get some opinions before I go any further.

reduckted:
Fixes #93 

Intellisense:
image

Docs:
image


madskristensen:
Awesome!! Thank you

PR #96, opened 2021-07-05
BaseCommand's Command and Package are always non-null

reduckted:
`BaseCommand.Command` and `BaseCommand.Package` can't be initialized when the object is created, but for all intents and purposes, those two properties are non-null because they are set immediately after the object is created in `BaseCommand.InitializeAsync()`.

For a consumer, needing to use the null-conditional operator for a property that will always be non-null is annoying, so it's best to declare them as non-nullable.

I'm not sure what the preferred way to do this is. I initially wrapped the properties in `#nullable disable`/`#nullable restore`, but decided to instead initialize them to `null!` (using the null-forgiving operator). I'm happy to go with whatever the recommended way is. 😄

### DIFF
--- a/src/Community.VisualStudio.Toolkit.Shared/Commands/BaseCommand.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/Commands/BaseCommand.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.ComponentModel.Design;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


src/Community.VisualStudio.Toolkit.Shared/Commands/BaseCommand.cs
```diff
@@ -96,7 +96,7 @@ protected virtual void Execute(object sender, EventArgs e)
                 {
                     await ex.LogAsync();
                 }
-            });
+            }).FireAndForget();
```
reduckted, line 99: Removing the null-conditional operator on this statement resulted in VSTHRD110 (_Observe the awaitable result of this method call by awaiting it, assigning to a variable, or passing it to another method._), hence why I've added the `FireAndForget()` call.

madskristensen:
Thanks

PR #99, opened 2021-07-05
Updated UI context casing

shaggygi:
Updated UI context casing throughout.  Appears to be the same in [docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivsmonitorselection.getcmduicontextcookie?view=visualstudiosdk-2019), as well.


madskristensen:
Thanks

PR #100, opened 2021-07-05
Use explicit types.

shaggygi:
Updated to use explicit types.  We've started to do this over in [.NET IoT](https://github.com/dotnet/iot) and I believe the Roslyn team is doing, as well.


madskristensen (commented):
This should be stated in the .editorconfig file, so VS will default to and enforce it going forward

madskristensen:
Awesome. Thank you

PR #101, opened 2021-07-05
Renamed NodeType to SolutionItemType

shaggygi:
Renamed NodeType to SolutionItemType.
Separated to own code file.
Updated related summaries and variables.


madskristensen:
Thanks

PR #105, opened 2021-07-07
Small fixes / improvements

bluetarpmedia:
- Fixed some spelling typos
- Corrected some comments
- Removed nullable IEnumerable<> return type from some methods that didn't need it
- Removed some nullable reference variable declarations when unneeded


madskristensen:
Thanks

PR #111, opened 2021-07-08
Add selected solution item to GetSelectedItemsAsync()

### LINKED ISSUE #107: Solution item not returned for GetSelectedHierarchiesAsync() when selecting items and the solution item
I could be misinterpreting how this should work, but seems like the solution item would also return when selecting along with other items.  Use image below as example.

image

✔️ : It returns the solution item and a count of 1 if I only select the solution item.
✔️ : It returns the folder item and a count of 1 if I only select the folder item.
✔️ : It returns both folder items (2nd folder not shown above) and a count of 2 if I select both folder items.
❌ : It returns the folder item and a count of 1 if I select the solution item and the folder item.  Seems like it would also return the solution item and a count of 2.

I've added other items (folders, projects, etc.) and get same results.  I believe the error occurs within [this if statement](https://github.com/VsixCommunity/Community.VisualStudio.Toolkit/blob/db3fc391d24a7670dbf8d5999af7ad32a2289656/src/Community.VisualStudio.Toolkit.Shared/Selection/Selection.cs#L105) as the `hierItem` is returned null for the solution item and doesn't get added to results.

shaggygi:
This fixes #107 (for the most part).  There are some other scenarios I noticed, but can be done in a future issue/PR.  For example, selecting multiple items under a project's Analyzers don't appear to return correctly.

This also updated some misspellings and found line variables swapped when showing a message box.


madskristensen:
Thanks

PR #114, opened 2021-07-09
Remove results.Distinct()

### LINKED ISSUE #109: O(n^2) implementation in Selection.GetSelectedHierarchiesAsync
This looks quadratic to me because `List.Contains` is O(n) and the `results` List grows in size to equal (or get close to) the size of `items`.

```
List<IVsHierarchyItem> results = new();

...

VSITEMSELECTION[] items = new VSITEMSELECTION[itemCount];
multiSelect.GetSelectedItems(0, itemCount, items);

foreach (VSITEMSELECTION item in items)
{
    IVsHierarchyItem? hierItem = await item.pHier.ToHierarcyItemAsync(item.itemid);
    if (hierItem != null && !results.Contains(hierItem))
    {
        results.Add(hierItem);
    }
}
```

Is the `results.Contains` necessary? Is there a scenario where `GetSelectedItems` returns duplicates?

If so, perhaps the method can add to `results` unconditionally and after the `foreach` completes, remove duplicates at the end.

bluetarpmedia:
Closes #109 

I believe @shaggygi's refactoring to use `results.Distinct` was to keep the previous logic of

```
if (hierItem != null && !results.Contains(hierItem))
```

But based on the discussion in #109 I don't think this is needed.

### DIFF
--- a/src/Community.VisualStudio.Toolkit.Shared/Selection/Selection.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/Selection/Selection.cs
@@ -163,7 +163,7 @@ namespace Community.VisualStudio.Toolkit
                 }
             }
 
-            return results.Distinct();
+            return results;
         }
     }
 }


shaggygi:
Yes, this is correct.  Mainly added until there was a fix on referenced issue.  Was not completely sure what the solution would be, so added short-term as a way to remove duplicates.

madskristensen:
Thansk

PR #123, opened 2021-07-13
*** POC *** Add Dependency Injection to the ToolkitPackage

StevenRasmussen:
*** This is a working POC that demonstrates the ability to integrate a DI container into the toolkit. Pertaining to discussion #122 ***

This adds a new interface `IToolkitServiceProvider` and implementation which wraps a `Microsoft.Extensions.DependencyInjection.IServiceProvider`.  This is added to the Visual Studio service collection so that you can retrieve this container using the regular VS methods, ie `VS.GetServiceAsync<SToolkitServiceProvider, IToolkitServiceProvider>`

~~While testing I noticed a slight inconvenience that would need to be documented: You have to add the `ProvdeCodeBase` attributes for the `Microsoft.Extensions.DependencyInjection.*` dlls to YOUR package instead of having them declared in the Toolkit package.  Maybe there's a workaround that I'm not aware of but these assembly level attributes aren't picked up by the implementing package and so the `*.pkgdef` does not include the appropriate references and Visual Studio throws an error if they are not defined in the implementing package.~~
This appears to be worked out by adding the `ProvideCodeBase` attributes to the `AssemblyInfo.cs` file.

Also, I was also able to use the most recent version (5.0.2) of the `Microsoft.Extensions.DependencyInjection` packages for v16 and v17 of the toolkit 😄. Versions 14 and 15 require v1.1.1 though.

~~Another potential issue that we may have is if there are multiple extensions installed that are based on the Toolkit. I'm not sure how Visual Studio would handle having multiple instances of the `SToolkitServiceProvider` registered in it's container.  Perhaps we could somehow register a generic service to the container that is based on the package class... something like:~~
This appears to work as I updated the PR to include this now:
```csharp
// Register in the VS container:
AddService(typeof(SToolkitServiceProvider<PackageTypeHere>), serviceCreatorCallback, true);

// Retrieve from the container:
IToolkitServiceProvider<PackageTypeHere> serviceProvider = await VS.GetRequiredServiceAsync<SToolkitServiceProvider<PackageTypeHere>, IToolkitServiceProvider>();
```

Another thought: We could also make this a completely separate package like `Community.VisualStudio.Toolkit.DependencyInjection` if we didn't want to pollute the main package with this.

### DIFF
--- a/src/Community.VisualStudio.Toolkit.Shared/DependencyInjectionContainerToolkitPackage.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/DependencyInjectionContainerToolkitPackage.cs
@@ -1,0 +1,55 @@
+﻿using System.Threading;
+using System;
+using Community.VisualStudio.Toolkit.Shared.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.VisualStudio.Shell;
+using Task = System.Threading.Tasks.Task;
+
+namespace Community.VisualStudio.Toolkit
+{
+    /// <summary>
+    /// Package that contains a DI service container.
+    /// </summary>
+    /// <typeparam name="TPackage"></typeparam>
+    [ProvideService(typeof(SToolkitServiceProvider<>), IsAsyncQueryable = true)]
+    public class DependencyInjectionContainerToolkitPackage<TPackage> : ToolkitPackage
+        where TPackage : AsyncPackage
+    {
+        /// <summary>
+        /// Custom ServiceProvider for the package.
+        /// </summary>
+        public IToolkitServiceProvider<TPackage> ServiceProvider { get; private set; } = null!; // This property is initialized in `InitializeAsync`, so it's never actually null.
+
+        /// <summary>
+        /// Initializes the <see cref="AsyncPackage"/>
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        /// <param name="progress"></param>
+        /// <returns></returns>
+        protected override async Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
+        {
+            await base.InitializeAsync(cancellationToken, progress);
+
+            ServiceCollection services = new ServiceCollection();
+            InitializeServices(services);
+            ServiceProvider = new ToolkitServiceProvider<TPackage>(services);
+
+            // Add the IToolkitServiceProvider to the VS IServiceProvider
+            AsyncServiceCreatorCallback serviceCreatorCallback = (sc, ct, t) =>
+            {
+                return Task.FromResult((object)this.ServiceProvider);
+            };
+
+            AddService(typeof(SToolkitServiceProvider<TPackage>), serviceCreatorCallback, true);
+        }
+
+        /// <summary>
+        /// Initialize the services in the DI container.
+        /// </summary>
+        /// <param name="services"></param>
+        protected virtual void InitializeServices(ServiceCollection services)
+        {
+            // Nothing
+        }
+    }
+}
--- a/src/AssemblyInfo.cs
+++ b/src/AssemblyInfo.cs
@@ -3,3 +3,5 @@
 using Microsoft.VisualStudio.Shell;
 
 [assembly: ProvideCodeBase(AssemblyName = "Community.VisualStudio.Toolkit")]
+[assembly: ProvideCodeBase(AssemblyName = "Microsoft.Extensions.DependencyInjection")]
+[assembly: ProvideCodeBase(AssemblyName = "Microsoft.Extensions.DependencyInjection.Abstractions")]
--- a/src/Community.VisualStudio.Toolkit.14.0/Community.VisualStudio.Toolkit.14.0.csproj
+++ b/src/Community.VisualStudio.Toolkit.14.0/Community.VisualStudio.Toolkit.14.0.csproj
@@ -10,6 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="Madskristensen.VisualStudio.SDK" Version="14.0.83" IncludeAssets="All" PrivateAssets="None" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="1.1.1" />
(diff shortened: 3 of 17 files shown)


src/Community.VisualStudio.Toolkit.Shared/DependencyInjectionContainerToolkitPackage.cs
```diff
@@ -0,0 +1,55 @@
...
+using Task = System.Threading.Tasks.Task;
+
+namespace Community.VisualStudio.Toolkit
+{
+    /// <summary>
+    /// Package that contains a DI service container.
+    /// </summary>
+    /// <typeparam name="TPackage"></typeparam>
+    [ProvideService(typeof(SToolkitServiceProvider<>), IsAsyncQueryable = true)]
+    public class DependencyInjectionContainerToolkitPackage<TPackage> : ToolkitPackage
```
StevenRasmussen, line 15: Someone can probably come up with a better name for this than I did 😛

yannduran: Maybe just `DIToolkitPackage` or `ToolkitPackageWithDI`?

CZEMacLeod:
> While testing I noticed a slight inconvenience that would need to be documented: You have to add the `ProvdeCodeBase` attributes for the `Microsoft.Extensions.DependencyInjection.*` dlls to YOUR package instead of having them declared in the Toolkit package. Maybe there's a workaround that I'm not aware of but these assembly level attributes aren't picked up by the implementing package and so the `*.pkgdef` does not include the appropriate references and Visual Studio throws an error if they are not defined in the implementing package. Also, I was also able to use the most recent version (5.0.2) of the `Microsoft.Extensions.DependencyInjection` packages for v16 and v17 of the toolkit 😄. Versions 14 and 15 require v1.1.1 though.

If you add the lines to [src/AssemblyInfo.cs](https://github.com/VsixCommunity/Community.VisualStudio.Toolkit/blob/master/src/AssemblyInfo.cs), they get 'injected' into the source code of the extension. You can see it already injects the ToolKit itself, and you could add the `Microsoft.Extensions.DependencyInjection.*` dlls here. The .cs file is injected into the Compile items list by the `Community.VisualStudio.Toolkit.props` file included in the package.

> Another thought: We could also make this a completely separate package like `Community.VisualStudio.Toolkit.DependencyInjection` if we didn't want to pollute the main package with this.

I really like DI - I have been using MEDI with Xamarin Forms in a not dissimilar way for some time, but I think perhaps this should be a separate package, that way you can omit the extra DLLs etc. if you are not using it, or use your own DI provider if you don't like the MEDI one.

StevenRasmussen:
@CZEMacLeod - Thanks for pointing that out!  I've updated accordingly. Re: separate package - the more I think about it the more I like it.  @madskristensen - I haven't seen you comment on the whole DI thing yet.  Is this something that you're interested in including into the project (whether in the main package or as a separate package)?  If not, I think I'll move forward with publishing this on my own.

StevenRasmussen:
> I don't think you need the CodeBase here

For some reason the `TestExtension` wouldn't work without it (and you'll notice that there is an entry for the toolkit itself there as well).  I think perhaps because the `Community.VisualStudio.Toolkit` is referenced as a project instead of a package in the `TestExtension` it doesn't work like it would if installed as a package.

yannduran:
I like the idea of a separate package too.

madskristensen:
Hey everybody. I just got back from vacation. I like the idea of DI, but I'm a fan of keeping it in a separate package. We want to make sure to keep the concept count of the toolkit as low as possible. I love the idea of additions like this to the toolkit though

StevenRasmussen:
Sounds good!  I'll work on a PR for this to make it a separate package... or possibly 2 packages.  I like the idea of making it extensible so that other DI containers could be used as well.  So I'd like to propose the following packages:
1) `Community.VisualStudio.Toolkit.DependencyInjection.Core` - contains all the core logic for wiring up the DI.
2) `Community.VisualStudio.Toolkit.DependencyInjection.Microsoft` - provides an implementation that utilizes the `Microsoft.Extensions.DependencyInjection.ServiceProvider`

yannduran:
Does `DependencyInjection` need to be spelled out. Wouldn't `DI` do?

CZEMacLeod:
> Does `DependencyInjection` need to be spelled out. Wouldn't `DI` do?

I think, like `Microsoft.Extensions.DependencyInjection.*`, it should have the full name.
With auto-complete and intellisense, I don't see that it makes a lot of difference in use.

I have to say though I'm not entirely sold on `Community.VisualStudio.Toolkit.DependencyInjection.Microsoft` as Microsoft have several DI technologies - MEF, MEF2, Unity and I'm sure there are others, as well as `Microsoft.Extensions.DependencyInjection`.
That said, I can't think of a better package name, other than perhaps `Community.VisualStudio.Toolkit.DependencyInjection.MEDI`, but I don't like that one either, and I'm absolutely sure we don't want `Community.VisualStudio.Toolkit.DependencyInjection.Microsoft.Extensions.DependencyInjection`. 🤣

yannduran:
Maybe we could start using `Community.VisualStudio.Extensions.xxx`, as opposed to the using `Toolkit` in he names, seeing as Mads would prefer to keep some distinction between the two (and I agree with that sentiment). 

So something like `Community.VisualStudio.Extensions.DependencyInjection.xxx` or just Community.VisualStudio.Extensions.DI.xxx. Just like `EF` isn't spelled out as `EntityFramework` in their namespaces. And something like `MEDI` is too terse, and even worse when added to `Community.VisualStudio.Toolkit.DependencyInjection`. I don't even know what it stands for.

But it's not just a matter of typing it, namespaces like `Community.VisualStudio.Toolkit.DependencyInjection.Microsoft` are just unnecessarily wordy. There's no need for it when anyone who knows what Dependency Injection is is going to know what DI is.

ErikEJ:
Fwiw, EF is spelled out in their name spaces.

StevenRasmussen:
Closing due to new PR #124

PR #125, opened 2021-07-17
Provide methods to register toolwindows and commands.

StevenRasmussen:
Adds the following extension methods for an `AsyncPackage`
1) `RegisterCommandsAsync` - scans the package assembly (and any other specified assemblies) for classes that inherit from the `BaseCommand<>` class and are decorated with the `CommandAttribute` and automatically calls the `InitializeAsync` method.
2) `RegisterToolWindows` - scans the package assembly (and any other specified assemblies) for classes that inherit from the `BaseToolWindow<>` class and automatically calls the `Initialize` method.


test/VSSDK.TestExtension/TestExtensionPackage.cs
```diff
@@ -25,18 +25,12 @@ public sealed class TestExtensionPackage : ToolkitPackage
         protected override async Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
         {
             // Tool windows
-            RunnerWindow.Initialize(this);
-            ThemeWindow.Initialize(this);
-            MultiInstanceWindow.Initialize(this);
+            this.AutoRegisterToolWindows();
```
madskristensen: Should this be renamed to `RegisterToolWindows` since it doesn't happen automatically? Same with the commands

madskristensen:
Should the `ToolkitPackage` do this automatically, so you don't need to call RegisterCommands manually?

StevenRasmussen:
> Should the `ToolkitPackage` do this automatically, so you don't need to call RegisterCommands manually?

I think it makes a lot of sense to do it by default!  As you said on the video on Friday, the toolkit should try to do things the best way by default and I think this falls under that category.  

I’m trying to think of a situation where you wouldn’t want the toolkit to automatically register a command/tool window but can’t think of one. Perhaps for commands we could add a property to the CommadAttrubute called ‘AutoRegister’ that defaults to true. That would provide the ability to turn it off for a command if you wanted to. Thoughts?

For tool windows we aren’t using an attribute and so we wouldn’t be able to control it that way.

bluetarpmedia:
> Perhaps for commands we could add a property to the CommandAttrubute called ‘AutoRegister’ that defaults to true. That would provide the ability to turn it off for a command if you wanted to. Thoughts?

I'd prefer to explicitly initialize commands and tool windows (whether with these new Register APIs or manually). So if the consensus is to go the auto-route I'd definitely like something like this to opt-out of it.

Here are a couple of cases where I think the opt-out is useful:
- Order of initialization (probably not great design, but perhaps a package needs to init something like a singleton first that the command/tool windows will need to access in their init)
- Conditional use of commands/tool windows (e.g. a commercial product where some commands are only enabled in licensed mode)

madskristensen:
@bluetarpmedia those are great points. Having two method calls in `InitalizeAsync` is pretty awesome already, so going the extra length might cause more confusion in some cases.

StevenRasmussen:
@bluetarpmedia - good points.  Sounds like the safer thing would be to keep it so that you manually call the methods.

madskristensen:
@StevenRasmussen is it ready to merge?

StevenRasmussen:
@madskristensen - Yes... sorry for the late merges. I noticed that I wasn't using explicit types but it should be ready now.  On that same note, if we want to use explicit types for the project perhaps we should update the '.editorconfig' to make using a `var` an error?  I personally prefer `var` everywhere, but I saw the PR get merged a couple of weeks back that made everything explicit and so I think consistency is more important at this point.  But if you're up for changing things back to `var` everywhere I'd be up for that conversation as well 😉

StevenBonePgh:
I'd prefer var as well. I used to be on the other side of this until a few years back when I realized how many extra changes were necessary to refactor to interfaces when introducing unit testing and DI to a 14+ year old codebase - a significant number of manual changes being 'cosmetic' replacements that would not have been necessary with var.  Whatever the decision, the .editorconfig should specify it.

madskristensen:
We're using explicit types to align with other .NET projects such as Roslyn. Yes, it should be added to .editorconfig. One cool thing about using explicit types is that you can then use the `new()` syntax.

madskristensen:
Thanks

StevenRasmussen:
> We're using explicit types to align with other .NET projects such as Roslyn. Yes, it should be added to .editorconfig. One cool thing about using explicit types is that you can then use the `new()` syntax.

Yeah... still not worth it in my opinion...
`var thing = new Thing();`
vs
`Thing = new();`
Potato Potatoe 😛 
I still just really like the readability and terseness of `var`, not to mention the benefits when refactoring as @StevenBonePgh mentioned.  Oh well... I have created a PR to update the .editorconfig as well as any remaining uses of `var`.

madskristensen:
I don't care either way, I just want consistency.

yannduran:
I'd definitely prefer to see `var` used everywhere! It's easy to hover over the `var` keyword to see the type, and there's even a function key you can hold down (`F1` maybe) that shows the types inline temporarily. Whereas using explicit types it makes for very ugly code and too much repetition.

StevenRasmussen:
> We're using explicit types to align with other .NET projects such as Roslyn.

FWIW - It appears that the AspNetCore team (which I think would arguably be the most used modern MS tech stack) uses `var` 😉

StevenBonePgh:
[Entity Framework Core](https://github.com/dotnet/efcore) also uses var.

PR #128, opened 2021-07-18
Custom style for PasswordBox

reduckted:
Relates to #106 

There are two changes to the toolkit library:
* A custom style for `PasswordBox` has been added, to make it look the same as a `TextBox`.
* The styles are now loaded from a XAML file, instead of being created from code in the `Themes` class.

I ran into similar problems to what @awschristou mentioned in #106, where the XAML styles could not be loaded in the regular way. There are a couple of problems with trying to load the XAML that way:
* The XAML file needs to reference certain namespaces, which also requires specifying the assembly (for example `clr-namespace:Microsoft.VisualStudio.Shell;assembly=Microsoft.VisualStudio.Shell.15.0`). The exact assembly differs between toolkit versions, so loading the XAML file directly from a resource using a XAML loader will fail for certain toolkit versions.
* XAML files in shared projects don't get any intellisense, which makes editing them _really_ painful.

Instead of defining the XAML file as a `Page` in the project file, I've defined it as an `EmbeddedResource`. This allows us to load the embedded resource, alter the namespaces/assemblies that are referenced in the XAML, then parse it into a `ResourceDictionary`. I've also added it as an `EmbeddedResource` directly to the project files (via `Common.props`) rather than in the shared project so that we can get intellisense when editing the file.

One other problem that I encountered is that there's no edit-and-continue for the XAML file. That makes editing the file and testing the changes a real pain, because you would have to make the change, start the experimental instance of Visual Studio, test the change you've made and if you need to make more changes, close the experimental instance and repeat the process.

To workaround this, when the toolkit is compiled in `DEBUG` mode, the `ResourceDictionary` is _not_ cached (so it's loaded each time the `UseVsTheme` property is enabled, and the file is loaded from disk instead of from the embedded resource. This means that if you've made changes to the XAML file, you can just toggle the _Use VSTheme_ checkbox in the _Theme Window_ and the new XAML is loaded and applied.

The test extension has had a couple of changes to the _Theme Window_ tool window:
* You can now change the Visual Studio theme from this tool window.
* Each styled control is shown with different states where applicable.
themes


madskristensen:
Awesome. Thanks!!

madskristensen:
Is `ThemeResources.xaml` not meant to be included in the project?

reduckted:
Not in the shared project, no. Intellisense doesn't work for XAML files in shared projects. It's included in each project via the `Common.props` file.

Issue #112, opened 2021-07-08
BaseOptionModel should serialize to the closest data type by default instead of binary


reduckted:
When saving a property that is a `bool`, the `BaseOptionModel` should store the value as a boolean using `WritableSettingsStore.SetBool()` instead of using a binary serializer and storing it as a string. 

The same should occur to `int`, `uint`, `long`, `ulong` and `string`.

madskristensen:
Why is this better than using serializing?

StevenBonePgh:
For one it is more performant to use the native type if possible. The second reason is that migrating an extension where the `WritableSettingsStore` was used in this manner to use `BaseOptionModel` instead, previously stored values of the specific type will fail to be read or deserialized. I added the `OverrideDataTypeAttribute` in #29 that can be applied to the property to specify the storage type, which may be a workaround. 

This suggestion, as proposed, would be a breaking change to anyone already using `BaseOptionModel` as-is that does not specify this attribute. Some built-in types, like byte and enumerations are edge cases for determining the closest native type. Then there is float, double, datetime, and custom types where current serialization would likely be the best. Could be made non-breaking if `BaseOptionModel ` is used as a base class and a new `BaseOptionModelEx` class could override the storage methodology - this would also allow for the `BaseOptionPage` to work with either implementation.

madskristensen:
I think we should only use a single method for storing the values. If that means using the right/closes data types, then let's get rid of the serialization completely. Would all other data types such as `DateTime` not just be stored as strings in this case?

madskristensen:
I'm not too concerned about breaking existing extensions on this one since the toolkit is still a -pre package. I'd rather we focus on getting it right. 

Will this change also mean that the attributes you added a while back could be removed?

StevenBonePgh:
I don't know of a reasonable way, other than the default serializer you already use, to handle custom or built in non-natively supported types. For commonly used types that are not exact matches we can hard-code how these are handled. I think for byte/short/ushort or enums with those underlying types we can promote to store as integer, char as string (or possibly short), datetime as long/Int64, and standard supported types as themselves, and for anything else we could use your serializer.

I'd suggest keeping the override attribute for porting from existing extensions where the type we use/infer by default is not the proper one, and for extensions like Tweakster that already use this you can apply the attribute to use the serializer to your existing properties.

I can give this one a go, perhaps a PR for it by Monday?

madskristensen:
[...]

PR #129, opened 2021-07-19
BaseOptionModel serializes to the closest data type by default (#112)

StevenBonePgh:
- Types supported: [Integral numeric types except for nint and nuint](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types), all enumeration types (via underlying type), bool, float, double, decimal, char, string, DateTime, DateTimeOffset, Color, and Guid.
- Types with caveats: string (null becomes string.empty), byte[] and MemoryStream which if null become empty byte arrays.
- Special Handling: Color, Guid, DateTime, DateTimeOffset, float, and double are handled specially to assure they round-trip. Convert.ChangeType is used for the remainder of the types, if the type is not already an exact match for the settings store.
- For other types, for example string[], the default serialization is now using XmlSerializer rather than BinarySerializer, which supports round-tripping null as empty string.
- Reflection on each property is performed once per options type (static) on the first load or save call. The reflection process creates open delegates that are used to get/set property values and the settings store.
- Retained ability to opt in to override collection name, property name, and storage type, including ability to use TypeConverter.
- Added unit test project, covering all mentioned above, using xUnit, Moq, and FluentAssertions.
- Added additional sample types to the TestExtension options (Color, enum, string[], string, DateTime)

As discussed, this is a breaking change, and settings stored with previous versions of the toolkit will log an error when Load is called for each setting, essentially reverting to the object's default values. Once Save is called, the native storage mechanism will overwrite the old settings with the proper type and no more errors will be seen.

### DIFF
--- a/Community.VisualStudio.Toolkit.sln
+++ b/Community.VisualStudio.Toolkit.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.31321.278
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31512.422
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Community.VisualStudio.Toolkit.16.0", "src\Community.VisualStudio.Toolkit.16.0\Community.VisualStudio.Toolkit.16.0.csproj", "{38DC82EA-91D6-4360-B76C-995708EE43A3}"
 EndProject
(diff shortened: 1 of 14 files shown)


madskristensen:
Thanks

madskristensen:
Thanks

madskristensen:
What is the purpose of `IOptionModelPropertyWrapper`?

StevenBonePgh:
> What is the purpose of IOptionModelPropertyWrapper
Force of habit/coding style for me, its almost automatic for me to do when coding to enable poylmorphic behavior without inheritance, DI, Mocking, etc.  Easy to remove if desired.
> Why not do a switch statement on the property type and save/load to the setting store using the appropriate type?
Most appropriate storage type is determined at construction of the wrapper and the delegates to the setting store and property are created then. Since we already know this, we don't need to switch on every possible type, perform the conversion, and call to the settings store (that code was long and ugly), we only need to handle the subset of types that require special consideration to enable them to roundtrip and/or are not supported by Convert.ToType/FromType. On load and save, the need to check property type is decreased based on the storage type - so no need to check the property type if there is no special conversion that applies to or from the storage type. On load, this also means when converting to the property type we can cast the storage value without the need to check types since we already know what it is.

PR #130, opened 2021-07-19
ProgressBar styles

reduckted:
Relates to #106.

There's a `Microsoft.VisualStudio.PlatformUI.SmoothProgressBar` which is already themed, so it seems like a better choice than a regular `ProgressBar`. No theming is needed, but the default height is thin (similar to what you see when searching in Solution Explorer), so I've added a resource that you can use to give it a thicker height if you want it to be like the progress bar in the `IVsThreadedWaitDialog`.


madskristensen:
Awesome

PR #131, opened 2021-07-19
Update .editorconfig to display errors when using the 'var' keyword instead of explicit types

StevenRasmussen:
All current uses of `var` were updated to use explicit types.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -73,9 +73,9 @@ dotnet_naming_style.start_underscore_style.required_prefix      = _
 # CSharp code style settings:
 [*.cs]
 # Prefer "var" everywhere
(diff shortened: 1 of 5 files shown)


madskristensen:
Thanks

yannduran:
Why? And why was there no discussion?

reduckted:
To enforce consistency. It was already a suggestion. This just enforces that code style. Given that there were only six code changes, most places were already following the convention.

mrlacey:
> Why? And why was there no discussion?

The discussion was had as part of https://github.com/VsixCommunity/Community.VisualStudio.Toolkit/pull/125

yannduran:
Thanks, I didn't see #125 until after I saw this.

PR #132, opened 2021-07-19
Small refactoring

bluetarpmedia:
Minor cleanup, removed some unneeded nullable variable declarations and merged some using statements.


madskristensen:
Thanks

PR #133, opened 2021-07-20
Included the type to cast to in the summary comment to make it appear in intellisense

reduckted:
The `<returns>` comment doesn't appear in intellisense, which makes it difficult to know what type to cast to for the service methods that return an `object`. I've moved that text into the `summary` element instead.

Before:
image

After:
image


madskristensen:
Awesome. Thanks

Issue #136, opened 2021-07-23
Determining Build Success/Fail


jcassidyav:
It would be great if the build done event passed along whether the build succeeded or failed.

i.e. the int fSuccess or is there another easy way to determine this from what is returned?

StevenRasmussen:
It looks like if you use the `VS.Build.Build*` methods that they return a `bool`, which I’m assuming is the result of the build. Looks like we could update the `Project.BuildAsync` method signature to return the result as well.

jcassidyav:
Could 
``` public event Action<Project?>? ProjectBuildDone;```
be changed to something like
 ```public event Action<ProjectBuildDoneResultArgs?>? ProjectBuildDone;```

where in ```IVsUpdateSolutionEvents2.UpdateProjectCfg_Done```

```ProjectBuildDone?.Invoke(new ProjectBuildDoneResultArgs(project, fSuccess==0));```

given something like

```

    public class ProjectBuildDoneResultArgs
    {
        public ProjectBuildDoneResultArgs(Project? project, bool isSuccesfull)
        {
            Project = project;
            IsSuccesfull = isSuccesfull;
        }

        public Project? Project { get; }
        public bool IsSuccesfull { get; }
    }
```

**Comment by shaggygi** [comment]
Noticed this thread a little late and seems like a good feature.

Understanding this is somewhat carried over and used in the VSSDK APIs, it seems like the wording should be "Completed" instead of "Done".

As an older family member tells me... "A cake gets done.  A chore gets completed."  😄 

So would it be worth renaming like...

```csharp
 public class ProjectBuildCompletedResultArgs
    {
        public ProjectBuildCompletedResultArgs(Project? project, bool isSuccesfull)
        {
            Project = project;
            IsSuccesfull = isSuccesfull;
        }

        public Project? Project { get; }
        public bool IsSuccesfull { get; }
    }
```

PR #138, opened 2021-07-26
Enrich output from build done events to include success/fail

jcassidyav:
Also fix solution event to report success/fail correctly.

Addresses #136


madskristensen:
Awesome. Thanks

PR #141, opened 2021-07-30
Add solutionFilename parameter to OnBeforeOpenSolution event.

pascalgyger:
Add the solutionFilename parameter for the OnBeforeOpenSolution event to the eventhandler. I need to know the path to the solution directory for this event and I believe this would be relevant for others that subscribe to this even too.


madskristensen:
Thanks

PR #150, opened 2021-08-19
Added GetDocumentAsync() on WindowsFrame and X# ProjectType

RobertvanderHulst:
Added GetDocumentAsync() on WindowsFrame to get the document opened in the frame

Also added the X# project type to the known ProjectTypes


madskristensen:
Thanks

madskristensen:
There was already a method called `GetDocumentViewAsync` on the WindowFrame class. Why did you add `GetDocumentAsync`?

RobertvanderHulst:
@madskristensen I am not sure when that method was created. In the nuget package that I was using this method did not exist.
You can delete the method if you want.

PR #159, opened 2021-08-30
Fix to Windows.GetAllWindowsAsync()

RobertvanderHulst:
The existing method GetAllWindowsAsync() was only returning ToolWindows.
I have renamed that method to make this clear.
I have also added a method GetAllDocumentWindowsAsync() and finally added method with the original name GetAllWindowsAsync that returns both the ToolWindows and Document Windows.


madskristensen:
Thanks

PR #168, opened 2021-09-10
Created a class for accessing and manipulating Solution Explorer

### LINKED ISSUE #98: Replacement for DTE2.ToolWindows.SolutionExplorer
I'm looking to see whether the Solution Explorer contains any items. See here:

https://github.com/reduckted/ProjectFilter/blob/b1e2681b41086fb3fba0c65774202bf593deb576/source/ProjectFilter/Services/SolutionLoadObserver.cs#L44-L55

What I'm doing might be a bit of a hack, but I'm executing a bunch of commands that operate on the Solution Explorer tool window to hide unloaded projects. If Solution Explorer hasn't been populated, then I wait a bit and try again. 

I also need to activate the Solution Explorer window ([see here](https://github.com/reduckted/ProjectFilter/blob/b1e2681b41086fb3fba0c65774202bf593deb576/source/ProjectFilter/Services/FilterService.cs#L323)) to allow those commands to work, and get the selection ([see here](https://github.com/reduckted/ProjectFilter/blob/b1e2681b41086fb3fba0c65774202bf593deb576/source/ProjectFilter/Services/FilterService.cs#L391)) so that I can restore the selected items afterwards.

I see there's an `IVsSolutionUIHierarchyWindow`, but it doesn't seem to provide access to the hierarchy in the window, and I can't find a corresponding `SVsSolutionUIHierarchyWindow` type.

reduckted:
Fixes #98.

The `SolutionExplorerWindow` class provides access to the Solution Explorer tool window, and lets you do things like setting the selection, expanding and collapsing items and applying filters.

The window can be retrieved via the `VS.Windows.GetSolutionExplorerWindowAsync()` method.

As well as the methods to manipulate the Solution Explorer window, I've also exposed the underlying `IVsUIHierarchyWindow` and `IVsSolutionUIHierarchyWindow` objects (they're actually the same object), as well as the `IVsWindowFrame` that contains Solution Explorer, in case you need to do stuff that the `SolutionExplorerWindow` doesn't provide (in my use case I needed access to the `IVsWindowFrame` so that I could activate the window).

----
To assist with testing, and to provide a demonstration, I've created one Solution Explorer filter and five commands in the test extension. These commands all appear in the _Extensions->VSSDK Test Extension->Solution Explorer_ menu.
image

The Solution Explorer filter also appears in the Solution Explorer toolbar.
image

* _Toggle VSIX Manifest Filter_ will enable and disable the filter. The filter will hide everything except for `*.vsixmanifest` files.
* _Select Current Project_ will get the current selection from Solution Explorer, then select the project that contains that selected file. This also works when multiple files are selected.
* _Edit Selected Item Label_ will have the same effect as pressing F2 in Solution Explorer.
* _Expand Selected Items_ will expand the selected items in Solution Explorer. There are multiple ways to expand an item, and they can be tested by holding down different keys when clicking this menu item.
    * Not pressing any keys will just expand the selected items (corresponds to `EXPANDFLAGS.EXPF_ExpandFolder`).
    * Holding down <kbd>Ctrl</kbd>will expand the selected items and all of their descendants (corresponds to `EXPANDFLAGS.EXPF_ExpandFolderRecursively`).
    * Holding down <kbd>Shift</kbd> will expand the ancestors of the selected items (corresponds to `EXPANDFLAGS.EXPF_ExpandParentsToShowItem`). To be able to demonstrate this, the selected items are first collapsed all the way up to the solution, and then they are expanded after a short pause.
    * Holding down both <kbd>Ctrl</kbd> and <kbd>Shift</kbd> will expand the ancestors and the descendants of the selected items.
* _Collapse Selected Items_ will collapse the selected items in Solution Explorer.

### DIFF
--- a/src/Community.VisualStudio.Toolkit.Shared/Solution/Solutions.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/Solution/Solutions.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
+using Task = System.Threading.Tasks.Task;
 
 namespace Community.VisualStudio.Toolkit
 {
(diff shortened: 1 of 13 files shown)


src/Community.VisualStudio.Toolkit.Shared/Solution/Solutions.cs
```diff
@@ -161,78 +162,84 @@ private async Task<IEnumerable<IVsHierarchyItem>> GetSelectedHierarchiesAsync()
             try
             {
                 svc.GetCurrentSelection(out hierPtr, out uint itemId, out IVsMultiItemSelect multiSelect, out containerPtr);
+                await AddHierarchiesFromSelectionAsync(hierPtr, itemId, multiSelect, results);
```
reduckted, line 165: ℹ️ The diff looks a bit weird here, but I've refactored the code that turns the result of `GetCurrentSelection` into `IVsHierarchy` objects into a separate method because `SolutionExplorerWindow.GetSelectionAsync` needed to do exactly the same thing. Hiding whitespace changes makes it a bit easier to see what changed.

madskristensen:
This is fantastic. Thank you so much!!

PR #171, opened 2021-09-18
Hide tool window

reduckted:
Related to #149 

I've created the `BaseToolWindow.HideAsync()` method which can be used to hide a tool window.

Example:
```cs
await RunnerWindow.HideAsync();
```


madskristensen:
Thanks

PR #173, opened 2021-09-25
Waited for the build to complete before returning from BuildProjectAsync and BuildSolutionAsync

### LINKED ISSUE #158: [BUG] .sqplproj Rebuild does not build
Build return OK, but no .dacpac file is built:

            var project = await GetProject(sqlprojPath);
            if (project == null) return null;

            var searchPath = Path.Combine(Path.GetDirectoryName(project.FullPath), "bin");

            var files = Directory.GetFiles(searchPath, "*.dacpac", SearchOption.AllDirectories);

            foreach (var file in files)
            {
                File.Delete(file);
            }

            var buildStartTime = DateTime.Now;

            var ok = await VS.Build.BuildProjectAsync(project, BuildAction.Rebuild);

            if (!ok)
            {
                throw new Exception("Dacpac build failed");
            }

            files = Directory.GetFiles(searchPath, "*.dacpac", SearchOption.AllDirectories);

reduckted:
Fixes #158

### DIFF
--- a/src/Community.VisualStudio.Toolkit.Shared/Build/Build.cs
+++ b/src/Community.VisualStudio.Toolkit.Shared/Build/Build.cs
@@ -33,7 +33,7 @@ namespace Community.VisualStudio.Toolkit
         }
 
         /// <summary>
-        /// Builds the solution or project if one is specified.
+        /// Builds the solution.
         /// </summary>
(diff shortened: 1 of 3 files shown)


madskristensen:
Awesome. Thank you!

PR #176, opened 2021-10-05
Used Directory.Build.props files to set build properties

### LINKED ISSUE #175: [Suggestion] Use Directory.Build.props files to manage build properties
With more projects being added to the solution, it's getting a bit difficult to determine where the build properties for a project are stored. Using `Directory.Build.props` files is a great convention that will help make it easier to specify build properties for specific types of projects.

You can read more about it here: https://docs.microsoft.com/visualstudio/msbuild/customize-your-build#directorybuildprops-and-directorybuildtargets

In order to really take advantage of `Directory.Build.props` files, we'll need to change the directory structure a bit. This is what I propose:
image

`/Directory.Build.props` (the one in the `Solution Items` folder) will define properties that are common to all projects in the solution. This would be things like `Nullable` and `LangVersion`.

`/src/Directory.Build.props` would define properties that are shared by projects that produce NuGet packages. This would be things like `Authors`, `Copyright`, SignAssembly`, etc.

`/src/toolkit/Directory.Build.props` would define properties that only apply to the toolkit projects. Things like `PackageID`, `RootNamespace`, `Reference` items, etc.

`/test/Directory.Build.props` would define properties for test projects.

reduckted:
Closes #175 

The new solution layout looks like this:
image

### `/Directory.Build.props`

This defines properties that are common to all projects, such as:
* `LangVersion`
* `Nullable`
* etc.

### `/src/Directory.Build.props`

This mainly defines properties that are common to all NuGet packages, like:
* `Authors`
* `SignAssembly`
* `AssemblyOriginatorKeyFile`
* etc.

This also defines the icon for NuGet packages, which means the analyzers NuGet package now has an icon.

### `/src/Directory.Build.targets`

The project files automatically include the `Directory.Build.props` files at the very start of the project (before any `PropertyGroups` are defined. This means the `MajorVersion` that is defined in each project file has not been set at the time that `Directory.Build.props` is included. Thankfully, the project files automatically include the `Directory.Build.targets` files at the very end of the project, so this is where the `MajorVersion` property can be used to build the complete `Version` property value.

### `/src/toolkit/Directory.Build.props`

This defines properties that are specific to the `Community.VisualStudio.Toolkit` projects, imports the shared project, and defines `ItemGroups` for references and "build" files that need to be included in the NuGet packages (which I have moved to the `src/toolkit/nuget` directory and used the same directory structure that they use within the NuGet package.

### `/src/toolkit/Directory.Build.targets`

This follows the same concept as the `/src/Directory.Build.targets` file - it uses the `MajorVersion` property, which is defined in each project, to define the `VSxx` compile-time constant that is used for conditional compilation.

### `/test/Directory.Build.props`

This sets properties that are common to all test projects, which at the moment is just `IsPackable = false`.

### Other Changes

I've moved `GenerateKnownCommands.ps1` to a `tools` directory to avoid clutter at the root of the repository, and I've moved the `VSSDK.TestExtension` project to the a `demo` folder.

~~I also noticed that I forgot to run the analyzer tests in the appveyor script (😨), so I've added them to the `test_script` section!~~


reduckted:
Oops. Looks like the analyzer unit tests never actually worked in appveyor (they do work locally, though). I'll remove them from the build script and fix them in a separate PR.

madskristensen:
This is awesome. Thank you!!

PR #177, opened 2021-10-06
Used the GUID for the new Find And Replace window in VS 2019+

### LINKED ISSUE #119: New Find and Replace WindowGuid
I came across the following while reviewing `WindowGuids`.   The "Find and Replace" window (shown below) appears to have a GUID {6324226F-61B6-4F28-92EE-18D4B5FE1E48}.  There are currently a few GUIDs related to Find, but I'm not sure what window they represent.  I also could not find a reference to this GUID in the docs.

Should this Find and Replace be added to `WindowGuids` and if so, what should the property name be?  FindReplace2?

image

reduckted:
Fixes #119


madskristensen:
Nice. Thanks

PR #180, opened 2021-10-09
Support for editing references

### LINKED ISSUE #164: [QUESTION] How to work with references in classic projects?
I used to be able to do this:
 
       public static void AddReference(Project project, string reference)
        {
            var vsProject = project.Object as VSProject;
            vsProject?.References.Add(reference);
        }

reduckted:
Closes #164.

This was a fun challenge (and a bit frustrating at times with the lack of documentation!).

From a `Project` you can now access its references using the `Project.References` property, which returns a `ReferenceCollection`.

You can add references using the `AddAsync()` method. You can add either assembly references or project references (either normal projects or shared projects):

```cs
string assemblyFileName = "...";
await project.References.AddAsync(assemblyFileName);

Project otherProject = (await VS.Solutions.GetAllProjectsAsync()).First(x => x.Name == "Other");
await project.References.AddAsync(otherProject);
```

The `ReferenceCollection` implements `IEnumerable<Reference>`, so you can iterate through its references.

```cs
foreach (Reference reference in project.References) {
    Debug.WriteLine(reference.Name);
}
```

`Reference` has two derived types for specific types of references: `AssemblyReference` and `ProjectReference`. Any other types of references are just a `Reference`.

```cs
var assemblyReferences = project.References.OfType<AssemblyReference>();
var projectReferences = project.References.OfType<ProjectReference>();
var otherReferences = project.References.Where(x => x.GetType() == typeof(Reference));
```

From an `AssemblyReference`, you can get the file name of the assembly. From a `ProjectReference`, you can get the `Project` object. All `Reference` types allow you to get the name of the reference and the underlying `IVsReference` object.

References can also be removed using the `RemoveAsync()` method.

```cs
// Remove all assembly references.
await project.References.RemoveAsync(project.References.OfType<ProjectReference>().ToArray());

// Remove the reference to the "Other" project.
var other = project.References.OfType<AssemblyReference>().First(x => x.Name == "Other");
await project.References.RemoveAsync(other);
```

### DIFF
--- a/src/toolkit/Community.VisualStudio.Toolkit.Shared/References/ProjectReference.cs
+++ b/src/toolkit/Community.VisualStudio.Toolkit.Shared/References/ProjectReference.cs
@@ -55,6 +55,8 @@ namespace Community.VisualStudio.Toolkit
 
         private Guid GetProjectGuid()
         {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
             // A `ProjectReference` can represent an `IVsProjectReference` or an `IVsSharedProjectReference`.
             if (VsReference is IVsProjectReference projectReference)
             {
--- a/src/toolkit/Community.VisualStudio.Toolkit.Shared/References/ReferenceCollection.cs
+++ b/src/toolkit/Community.VisualStudio.Toolkit.Shared/References/ReferenceCollection.cs
@@ -238,6 +238,7 @@ namespace Community.VisualStudio.Toolkit
 
         IEnumerator IEnumerable.GetEnumerator()
         {
+            ThreadHelper.ThrowIfNotOnUIThread();
             return GetEnumerator();
         }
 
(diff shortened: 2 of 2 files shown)


madskristensen:
This is amazing. Thank you so much!!

SFC-Sarge:
@reduckted; How do you get the current project references from the active project. I am working on @madskristensen [HelpExplorer](https://github.com/madskristensen/HelpExplorer) project using SDK Style projects and I only get null returned from ActiveProject.Reference.  We would like to be able to easily and quickly return what type of project the developer is currently working on. We can get the base line project from the capabilities, but for things like (xUnit, MSTest, NUnit) or (WPF User Control, WPF Custom Control, or WPF Library) all the provided capabilities for those are the same, respectively. To provide the new or beginning developer with helpfully URL links for what they are working on, it would be better to provide a link for xUnit specific, if that is what they are working on, rather than a general testing URL that may or may not provide any help to the developer.
This line returns a location for the project.assets.json file and would work to determine xUnit, MSTest, or NUnit for the currenly selected active project. 
( _`var projectAssetsFile = await activeProject.GetAttributeAsync("ProjectAssetsFile");`_) 
but since the project.assets.json doesn't list anything about (WPF User Control, WPF Custom Control, or WPF Library) it will not work.

SFC-Sarge:
@reduckted; Is there a way to get the baseline template from VS. Solution that was used to create the current project. (i.e. When the user created the project from a Visual Studio installed templates.) What was the template name and template description since somewhere in the data is the type of project it would create and is currently being used by the selected active project?

PR #181, opened 2021-10-10
Fix analyzer tests

reduckted:
The problem with the analyzer tests was the line endings were different to what was expected. I suspect that the repository is checked out with LF line endings in Appveyor, but because it's running on Windows, Roslyn uses CRLF for any edits that it makes during the code fix. That caused the output of the code fix to be different to what was defined in the test.


madskristensen:
Thanks

Issue #184, opened 2021-10-18
Community.VisualStudio.Toolkit.props includes cs files in all projects, regardless of their types


RobertvanderHulst:
In the Community.VisualStudio.Toolkit.props file there is a line

`<Compile Include="$(MSBuildThisFileDirectory)*.cs" />`

This should only be done when this file is included in a C# project.

I think a condition should be added:

`Condition="'$(Language)'=='C#'"`

PR #185, opened 2021-10-18
Added condition to only include the .cs files for C# style projects

RobertvanderHulst:
When you include the community project in a non c# project then this lines causes the build for that project to fail.
With the condition this is no longer a problem


RobertvanderHulst:
This fixed #184

madskristensen:
Great. Thanks

PR #195, opened 2021-10-29
Build target to check the version of Newtonsoft.Json

### LINKED ISSUE #190: Build-time check for Newtonsoft.Json version compatibility
Specific versions of Visual Studio use specific versions of `Newtonsoft.Json` (see [this blog post](https://devblogs.microsoft.com/visualstudio/using-newtonsoft-json-in-a-visual-studio-extension/) for more details). It would be great if we had some sort of compile-time check that would raise a build error if the version of `Newtonsoft.Json` that is used in the project is greater than the version that Visual Studio will use.

For example, if you've installed `Community.VisualStudio.Toolkit` v15 in your project, then you are targeting Visual Studio v15. If the project also has `Newtonsoft.Json` installed, then the version should be less than or equal to 8.0.3.

reduckted:
Fixes #190 

This adds a compile-time check that the version of `Newtonsoft.Json` that is referenced by your project is no higher than the version used by the corresponding version of Visual Studio. If the version of `Newtonsoft.Json` is less that or equal to the version used by Visual Studio (or you don't reference `Newtonsoft.Json`) then nothing is reported. If the version is greater, then a warning is generated.

json

### How it works

1. The highest version of `Newtonsoft.Json` that can be used is defined in four separate `.props` files - one for each version of the toolkit.
2. When the NuGet package is created, the `.props` file that corresponds to the major version of Visual Studio is included in the NuGet package. That `.props` file is imported by the `Community.VisualStudio.Toolkit.props` file that is also included in the NuGet package.
3. When consuming the NuGet package, a build target runs after references are resolved that finds the version of `Newtonsoft.Json` that is used.
4. If that version is greater that the version specified in the `.props` file from this NuGet package, then a warning is generated.

### Limitations

* This only works with `PackageReference`-style NuGet references. The `packages.config`-style NuGet references are not supported, but I can't imaging that being an issue these days.

### Changes
* To make the `Community.VisualStudio.Toolkit.targets` file less cluttered, I've refactored the existing Extensibility Essentials extension check into its own `.targets` file and imported it into `Community.VisualStudio.Toolkit.targets`. Likewise, this new check is defined in its own `.targets` file and imported the same way.
* I've renamed some of the build properties that were used in the Extensibility Essentials extension check that are intended to be "private" (i.e. used only within that build target). They now start with a `CVST_` prefix (short for **C**ommunity **V**isual **S**tudio **T**oolkit). I did this to avoid the possibility of overwriting/replacing build properties that come from other parts of the build process. I think it's pretty unlikely that would occur, but names like `MSBuildMajorVersion` were a bit too generic for my liking, so I thought I'd play it safe. Properties that are intended to be "public" to allow the targets to change their behavior (such as `ExtensibilityEssentialsLevel`) don't have that prefix.
* I've changed the error/warning codes that can be produced by the build from `VSTKxxx` to `CVSTBLDxxx`. Analyzer warnings use the `CVSTxxx` format, so I felt that `CVSTBLDxxx` was a better choice.


reduckted:
🤔 I hadn't thought about that, but I think that could still allow you to install the wrong version. [`Microsoft.VisualStudio.SDK` v16.0.206](https://www.nuget.org/packages/Microsoft.VisualStudio.SDK/16.0.206) uses a version range of `>= 9.0.1` for `Newtonsoft.Json`, which means you would be able to install v13.0.1.

PR #199, opened 2021-11-03
Improved unloaded project support

### LINKED ISSUE #191: Better support for unloaded projects
Dealing with unloaded projects is a bit difficult at the moment. 

* `Solutions.GetAllProjectsAsync()` only returns loaded projects.
* `Solutions.GetAllProjectHierarchiesAsync()` only returns loaded projects.
* `IVsSolution.GetAllProjectHierarchys()` (an extension method) allows a `__VSENUMPROJFLAGS` value to be specified which can be used to include unloaded projects (though the `__VSENUMPROJFLAGS` enum isn't very intuitive in my opinion).
* When trying to turn an `IVsHierarchy` into a `SolutionItem`, the item type always ends up as `Unknown`.

reduckted:
Fixes #191.

Changes:

* `ProjectStateFilter` is an enum that lets you choose between getting loaded, unloaded or all projects.
* `Solutions.GetAllProjectHierarchiesAsync` and `Solutions.GetAllProjectsAsync` now have an optional `ProjectStateFilter` property. The default is `Loaded` to match the previous behavior.
* `IVsSolutionExtensions.GetAllProjectHierarchies` also takes a `ProjectStateFilter`, which is converted to the corresponding `__VSENUMPROJFLAGS` value. The overload that takes a `__VSENUMPROJFLAGS` still exists.
* When getting the type for a `SolutionItem`, a "stub hierarchy" is treated as a project.

### DIFF
--- a/src/toolkit/Community.VisualStudio.Toolkit.Shared/ExtensionMethods/IVsSolutionExtensions.cs
+++ b/src/toolkit/Community.VisualStudio.Toolkit.Shared/ExtensionMethods/IVsSolutionExtensions.cs
@@ -37,16 +37,28 @@ namespace Microsoft.VisualStudio.Shell.Interop
         /// <summary>
         /// Gets all projects in the solution as IVsHierarchy items.
         /// </summary>
-        public static IEnumerable<IVsHierarchy> GetAllProjectHierarchys(this IVsSolution solution)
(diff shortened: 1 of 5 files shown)


madskristensen:
Awesome. Thanks

PR #200, opened 2021-11-04
Made SolutionItem.Name and FullPath read-only

### LINKED ISSUE #198: Should SolutionItem.Name be read-only?
The `Name` property of `SolutionItem` is read-write, but setting it has no effect on the underlying `IVsHierarchy` that the  `SolutionItem` wraps. Should this property be read-only instead?

`FullPath` has the same issue.

https://github.com/VsixCommunity/Community.VisualStudio.Toolkit/blob/e5230bb3766367e8cccab6f3e3471ec63d7ea4df/src/toolkit/Community.VisualStudio.Toolkit.Shared/Solution/SolutionItem.cs#L41-L49

reduckted:
I've also changed `Name` to return the `IVsHierarchy.CanonicalName` property rather than the `Text` property. I've added a `Text` property to get the hierarchy text.

The only difference I've been able to find between `CanonicalName` and `Text` is when a project is unloaded. For an unloaded project the `CanonicalName` will be "ProjectName", but the `Text` will be "ProjectName (unloaded)". 

I'm happy to revert the `Name` property back to returning the `IVsHierarcny.Text` if you think that would be better.

Fixes #198.


madskristensen:
Awesome. Thanks!!

PR #202, opened 2021-11-06
Used GetHierarchyItem instead of TryGetHierarchyItem

### LINKED ISSUE #201: Should ToHierarchyItemAsync always create the IVsHierarchyItem?
`ToHierarchyItemAsync` calls `IVsHierarchyItemManager.TryGetHierarchyItem`. The [documentation](https://docs.microsoft.com/dotnet/api/microsoft.visualstudio.shell.ivshierarchyitemmanager.trygethierarchyitem) states:

> Gets a cached `IVsHierarchyItem` representing the specified hierarchy and item identifier, **if one has already been created.**

The problem with that is `ToHierarchyItemAsync` ends up returning null if the project has not been displayed in Solution Explorer (for example, if it's an unloaded project).

`IVsHierarchyItemManager` has a `GetHierarchyItem` method that:

> Gets or creates an `IVsHierarchyItem` corresponding to the specified hierarchy and item identifier.

ℹ️ Note that `GetHierarchyItem` doesn't return null (at least as far as my testing has shown).

I was considering adding a new method that uses `GetHierarchyItem` instead, or perhaps add an optional parameter to `ToHierarchyItemAsync`, but I'm now wondering if we should _always_ be using `GetHierarchyItem` instead of `TryGetHierarchyItem`.

I can think of two options:

1. Change `ToHierarchyItemAsync` to always call `GetHierarchyItem`. We could possibly add a new method to get the hierarchy item only if it has already been created in case that's actually something that people need to do.
```cs
public static async Task<IVsHierarchyItem> ToHierarchyItemAsync(this IVsHierarchy hierarchy, uint itemId);

public static async Task<IVsHierarchyItem?> TryConvertToHierarchyItemAsync(this IVsHierarchy hierarchy, uint itemId);
```

2. Leave `ToHierarchyItemAsync` as it is and create a new method that calls `GetHierarchyItem`.
```cs
public static async Task<IVsHierarchyItem?> ToHierarchyItemAsync(this IVsHierarchy hierarchy, uint itemId);

public static async Task<IVsHierarchyItem> GetOrCreateHierarchyItemAsync(this IVsHierarchy hierarchy, uint itemId);
```

Any thoughts? 

In case it's of any use, searching across GitHub for both methods shows that `GetHierarchyItem` is used more, but that could be skewed by forks.
* `GetHierarchyItem`: [148 results](https://github.com/search?l=C%23&p=2&q=GetHierarchyItem&type=Code)
* `TryGetHierarchyItem`: [34 results](https://github.com/search?l=C%23&q=TryGetHierarchyItem&type=Code)

reduckted:
Fixes #201

### DIFF
--- a/src/toolkit/Community.VisualStudio.Toolkit.Shared/ExtensionMethods/IVsHierarchyExtensions.cs
+++ b/src/toolkit/Community.VisualStudio.Toolkit.Shared/ExtensionMethods/IVsHierarchyExtensions.cs
@@ -24,35 +24,29 @@ namespace Microsoft.VisualStudio.Shell.Interop
         /// <summary>
         /// Converts a <see cref="IVsHierarchy"/> to a <see cref="IVsHierarchyItem"/>.
         /// </summary>
-        /// <returns>Returns <see langword="null"/> if unable to get the hierarchy item.</returns>
-        public static async Task<IVsHierarchyItem?> ToHierarchyItemAsync(this IVsHierarchy hierarchy, uint itemId)
+        public static async Task<IVsHierarchyItem> ToHierarchyItemAsync(this IVsHierarchy hierarchy, uint itemId)
         {
             if (hierarchy == null)
             {
-                return null;
+                throw new ArgumentNullException(nameof(hierarchy));
             }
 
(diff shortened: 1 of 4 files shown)


madskristensen:
Thanks

PR #206, opened 2021-11-09
Fixes #205

### LINKED ISSUE #205: Several Solution events get null objects
Several of the solution events, such as OnAfterOpenProject() and OnAfterLoadProject() do not pass an null for their parameters.
I am not sure if the code that gets the project with SolutionItem.FromHierarchy(pHierarchy, 1) is wrong, or if there is another issue.

I also noticed that a null check inside OnBeforeUnloadProject() checks for **OnBeforeCloseProject**  instead of **OnBeforeUnloadProject**:
```
int IVsSolutionEvents.OnBeforeUnloadProject(IVsHierarchy pRealHierarchy, IVsHierarchy pStubHierarchy)
        {
            ThreadHelper.ThrowIfNotOnUIThread();

            if (OnBeforeCloseProject != null)
            {
                Project? project = SolutionItem.FromHierarchy(pRealHierarchy, 1) as Project;
                OnBeforeUnloadProject?.Invoke(project);
            }
            return VSConstants.S_OK;
        }

```

RobertvanderHulst:
This fixes the following issues:
- incorrect ItemId was used to retrieve the project object
- Incorrect check for OnBeforeUnloadProject event handler

### DIFF
--- a/demo/VSSDK.TestExtension/TestExtensionPackage.cs
+++ b/demo/VSSDK.TestExtension/TestExtensionPackage.cs
@@ -37,6 +37,20 @@ namespace VSSDK.TestExtension
             VS.Events.DocumentEvents.BeforeDocumentWindowShow += DocumentEvents_BeforeDocumentWindowShow;
             VS.Events.ProjectItemsEvents.AfterRenameProjectItems += ProjectItemsEvents_AfterRenameProjectItems;
             VS.Events.ProjectItemsEvents.AfterRemoveProjectItems += ProjectItemsEvents_AfterRemoveProjectItems;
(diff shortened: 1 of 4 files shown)


reduckted:
Nice!

`OnBeforeCloseProject` is also using `1` as the item ID. Should that be changed as well?

https://github.com/VsixCommunity/Community.VisualStudio.Toolkit/blob/39c4a05c127aaf8921fa266f2d1383de8cb9b6ff/src/toolkit/Community.VisualStudio.Toolkit.Shared/Solution/SolutionEvents.cs#L102-L111

RobertvanderHulst:
Yes, I have overlooked that one.

madskristensen:
Awesome. Thanks so much!!

Issue #221, opened 2021-11-15
ProjectTypes.CSHARP GUID is incorrect


RFBomb:
I Believe C# GUID is incorrect. I can submit a PR if you want me to.

This is my use case:
https://docs.microsoft.com/en-us/visualstudio/extensibility/internals/registering-single-file-generators?view=vs-2022

I pulled the working GUID from their example.

This code fails to register my SingleFileGenerator for use:
`[CodeGeneratorRegistration(typeof(XSDCustomTool), "XSDCustomTool", ProjectTypes.CSHARP, GeneratesDesignTimeSource = true)]`

This code registers succesfully:
`[CodeGeneratorRegistration(typeof(XSDCustomTool), "XSDCustomTool", "{FAE04EC1-301F-11D3-BF4B-00C04F79EFBC}", GeneratesDesignTimeSource = true)]`

also, the constants `WINDOWS_CSHARP ` and `CSHARP ` are identical.

Not Working:__ "{FAE04EC**0**-301F-11D3-BF4B-00C04F79EFBC}"
Working:______  "{FAE04EC**1**-301F-11D3-BF4B-00C04F79EFBC}"

As for VB, I didn't see the example GUID from MS listed at all on in the ProjectTypes files.
`"{164b10b9-b200-11d0-8c61-00a0c91e29d5}"`

To me, this suggests that these GUIDs differ somehow from where the "ProjectTypes" guids were gotten from. Maybe these should have their own static class?

edit:
Here are the registry keys I got from looking at my registry for a VS2019 install, inside of 
`HKEY_CURRENT_USER\Software\Microsoft\VisualStudio\10.0_Config\Generators`

VB -> {164B10B9-B200-11D0-8C61-00A0C91E29D5}
J# -> {e6fdf8b0-f3d1-11d4-8576-0002a516ece8}
C# -> {FAE04EC1-301F-11D3-BF4B-00C04F79EFBC}
LinqToSQL -> {39C9C826-8EF8-4079-8C95-428F5B1C323F}
Unknown -> {349C5850-65DF-11DA-9384-00065B846F21}

madskristensen:
Yeah, the C# guid might be incorrect as you point out. Let's fix it

RFBomb:
I was digging through my registry and searching for other GUIDs that SingleFileGenerators will use. I had edited my original post just as you were commenting, so unsure if you saw it. 

[...]

PR #222, opened 2021-11-15
Create SingleFileGeneratorTypes.cs

RFBomb:
ProjectTypes.cs contains the GUIDs for the ProjectTypes.
SingleFileGenerators do not use the same GUIDs, so a new static class was produced to provide those GUIDs as they become known.

### DIFF
--- a/src/toolkit/Community.VisualStudio.Toolkit.Shared/Helpers/SingleFileGeneratorTypes.cs
+++ b/src/toolkit/Community.VisualStudio.Toolkit.Shared/Helpers/SingleFileGeneratorTypes.cs
@@ -24,3 +24,4 @@ namespace Community.VisualStudio.Toolkit
         public const string VisualBasic = "{164B10B9-B200-11D0-8C61-00A0C91E29D5}";
     }
 }
+
--- a/src/toolkit/Community.VisualStudio.Toolkit.Shared/Helpers/SingleFileGeneratorTypes.cs
+++ b/src/toolkit/Community.VisualStudio.Toolkit.Shared/Helpers/SingleFileGeneratorTypes.cs
@@ -1,0 +1,26 @@
+namespace Community.VisualStudio.Toolkit
+{
+    /// <summary>
+    /// SingleFileGenerators rely on Language GUIDs rather than ProjectType GUIDs. <br/>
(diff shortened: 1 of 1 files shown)


madskristensen:
Nice. thank you

src/toolkit/Community.VisualStudio.Toolkit.Shared/Helpers/SingleFileGeneratorTypes.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    /// <remarks>
+    /// <see href="https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivssinglefilegenerator?view=visualstudiosdk-2019"/>
+    /// <para> Example class Attribute Usage: <br/>
+    /// [Guid("Guid-For-Your-IVsSingleFileGenerator")] <br/>
+    /// [ComVisible(true)] <br/>
+    /// [ProvideObject(typeof(YourCustomToolClass))] <br/>
+    /// [CodeGeneratorRegistration(typeof(YourCustomToolClass), "NameOfCustomTool", SingleFileGeneratorTypes.CSHARP, GeneratesDesignTimeSource = true, GeneratorRegKeyName = "CustomTool for C#")] <br/>
+    /// </remarks>
+    public static class SingleFileGeneratorTypes
```
StevenRasmussen, line 15: If these indeed are based on a 'Language' then I might suggest naming this something like `LanguageTypes`.  That being said, I don't know enough about these GUID's and where they all might be used so perhaps this name makes sense as-is.

RFBomb: So, the reason I named it like that is because the only reference I've been able to find for those GUIDS was specifically for SingleFileGenerators, such as the MSDataSetGenerator microsoft provides for XSD files. 

In my original issue, #221 , I had also shown where I got those GUIDs from: the 'Generators' registry key. 

Now, it _may_ make sense for it to be renamed to LanguageTypes, I don't disagree there. But I kinda figured that that would already be synonymous to the ProjectType constant class, since your project is compiling a specific language anyway.

PR #223, opened 2021-11-15
Update SingleFileGeneratorTypes.cs

RFBomb:
I thought that adding the GeneratorRegKeyName would add a description to the registry when registering. Instead it broke it completely. (Debugging in test instance it worked, but when the vsix was installed manually it failed to register.)

I'm removing the suggestion to have it in there since it doesn't work with that segment installed.


madskristensen:
Yikes, nice catch. Thansk

PR #238, opened 2021-12-07
Added the ability to show an Info Bar on the main window

### LINKED ISSUE #231: Create Info Bar on the main window
It seems like an Info Bar can only be created on a tool window or document window using the `InfoBarFactory` class.

An Info Bar can be added to the main window using:
```cs
IVsShell shell;
shell.GetProperty(__VSSPROPID7.VSSPROPID_MainWindowInfoBarHost, out var host);
return new InfoBar((IVsInfoBarHost)host, model);
```

Can this be achieved using the current overloads of `InfoBarFactory.CreateAsync()`? For example, is there a window GUID or `IVsWindowFrame` that represents the main window. Or do we need to add a new overload?

reduckted:
Fixes #231

```cs
InfoBar bar = await VS.InfoBar.CreateAsync(new InfoBarModel("Hello World!", KnownMonikers.Services, true));
await bar.TryShowInfoBarUIAsync();
```

image

### DIFF
--- a/src/toolkit/Community.VisualStudio.Toolkit.Shared/Notifications/InfoBar.cs
+++ b/src/toolkit/Community.VisualStudio.Toolkit.Shared/Notifications/InfoBar.cs
@@ -11,6 +11,25 @@ namespace Community.VisualStudio.Toolkit
     public class InfoBarFactory
     {
         /// <summary>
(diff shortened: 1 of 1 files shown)


madskristensen:
Thanks

PR #239, opened 2021-12-07
Fixed showing an Info Bar on a window via the window GUID

reduckted:
`GetFrameFromIdentifierAsync` always returned null, so `CreateAsync(string, InfoBarModel)` would never find the `IVsWindowFrame` and therefore would never create the `InfoBar`.


madskristensen:
Awesome. Thank you!!

PR #240, opened 2021-12-07
Methods to load and unload projects

reduckted:
The `Project` class now has some additional methods to load and unload projects.

Load a project:
```cs
await project.LoadAsync();
```

Unload a project:
```cs
await project.UnloadAsync();
```

Check if a project is loaded:
```cs
if (project.IsLoaded) { }
```

Loading or unloading a project seems to dispose of the underlying `IVsHierarchy` object, so I've added an internal `SolutionItem.Update()` method that can be used to give the SolutionItem a new `IVsHierarchyItem`.

I've also added a couple of commands to the demo extension to demonstrate the new functionality.


madskristensen:
Fantastic. Thanks!!

PR #242, opened 2021-12-10
Example of adding a toolbar to a tool window

reduckted:
This demonstrates how to add a toolbar to a tool window. 

Documentation for how to do this _without_ using the toolkit can be found here: https://docs.microsoft.com/en-us/visualstudio/extensibility/adding-a-toolbar-to-a-tool-window?view=vs-2022

Doing this _with_ the toolkit is basically the same. The toolbar is added by setting the `Toolbar` property on the internal `Pane` class of your tool window, because that's the class that inherits `ToolWindowPane`.

In this example, I've used a [`CommandPlacement`](https://docs.microsoft.com/en-us/visualstudio/extensibility/commandplacements-element?view=vs-2022) in the `.vsct` file to add the "Build Solution Async" command to the toolbar. The command also continues to appear in the "VSSDK Test Extension" menu. 

I've also added a new command to demonstrate one way that you can communicate between the toolbar and the tool window. As I've stressed in the comments, this should _not_ be considered a "best practice" or a "recommended" way to achieve that communication, but nor is it the wrong way. It's simply demonstrating that it is possible for a toolbar button to send messages to the tool window. 😄 

image

### DIFF
--- a/demo/VSSDK.TestExtension/Commands/SendMessageToRunnerWindowCommand.cs
+++ b/demo/VSSDK.TestExtension/Commands/SendMessageToRunnerWindowCommand.cs
@@ -1,0 +1,21 @@
+﻿using System.Threading.Tasks;
+using Community.VisualStudio.Toolkit;
+using Microsoft.VisualStudio.Shell;
+
+namespace TestExtension
+{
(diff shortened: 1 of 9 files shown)


madskristensen:
Awesome. Thanks

PR #244, opened 2021-12-23
Analyzer for adding ComVisibleAttribute to DialogPage

reduckted:
This pull request introduces a new code analyzer. This analyzer will detect classes that inherit from `DialogPage` that do not have the `[ComVisible(true)]` attribute. That attribute is required if the class `DialogPage` needs to support profiles (and there's no harm adding it if profiles don't need to be supported). Without it, exporting settings will silently ignore the settings defined in the `DialogPage`.


src/analyzers/Community.VisualStudio.Toolkit.Analyzers/Analyzers/CVST001CastInteropServicesCodeFixProvider.cs
```diff
@@ -208,58 +208,5 @@ private static bool IsDeclaratorForObject(SyntaxNode node, out VariableDeclarati
             variableDeclaration = null!;
             return false;
         }
-
-        private static SyntaxList<UsingDirectiveSyntax> AddUsingDirectiveIfMissing(SyntaxList<UsingDirectiveSyntax> usings, NameSyntax namespaceName)
```
reduckted, line 212: ℹ️ I've moved this method into the new `CodeFixProviderBase` class.

madskristensen:
Awesome. Thank you!!

PR #245, opened 2021-12-29
Hide code files that are included from the NuGet package

reduckted:
The toolkit includes an AssemblyInfo.cs file which ensures that the `ProvideCodeBase` attribute is applied to your extension. This file shows up in Solution Explorer, which is a bit confusing because it's not a file that you should edit.

image

I've added `Visible=false` metadata to the `Compile` item so that the file doesn't appear in Solution Explorer but is still included in the compilation.


madskristensen:
Excellent. Thank you!!

PR #267, opened 2022-01-18
MSBuild target for publishing to Visual Studio Marketplace

reduckted:
Related to #256.

This adds a build target called `PublishToMarketplace`. It uses the [VsixPublisher.exe](https://docs.microsoft.com/visualstudio/extensibility/walkthrough-publishing-a-visual-studio-extension-via-command-line) that comes with the `Microsoft.VSSDK.BuildTools` NuGet package.

You can publish an extension to the marketplace by running this
```
msbuild path/to/MyExtension.csproj /t:PublishToMarketplace /p:Configuration=Release /p:PersonalAccessToken=secret
```

~~The build target depends on `Rebuild`, which means a full rebuild of the project will occur before publishing. That means you don't need to run the `Build` target beforehand.~~

You can have the extension built before publishing by specifying the `BuildBeforePulish` property:
```
msbuild path/to/MyExtension.csproj /t:PublishToMarketplace ...snip... /p:BuildBeforePublish=true
```

The `publish.json` manifest file will be found by walking up the directory tree from the project directory. If you want to specify a specific manifest, you can specify the path to the file:
```
msbuild path/to/MyExtension.csproj /t:PublishToMarketplace ...snip... /p:PublishManifest=path/to/the/manifest.json
```

If you've moved the extension file somewhere else, you can override the file path by specifying the `PublishExtension` property:
```
msbuild path/to/MyExtension.csproj /t:PublishToMarketplace ...snip... /p:PublishExtension=path/to/the/extension.vsix
```

If you want to ignore specific warnings from `VsixPublisher.exe`, you can specify them as a CSV string:
```
msbuild path/to/MyExtension.csproj /t:PublishToMarketplace ...snip... /p:PublishIgnoreWarnings=warning01,warning02
```


madskristensen:
When publishing from a CI pipeline, the build and test runs have already happened by the time it will publish to the Marketplace. Does it make sense to make the rebuild optional? Also, what if the output .vsix was placed in a separate location, can the publishing task grab it from there instead of the path specified in the csproj?

reduckted:
@madskristensen Yep, that makes sense. I'll make a change that makes the rebuild opt-in instead of opt-out.

The extension path can be overridden by specifying the `PublishExtension` property (though I just spotted a bug in it that I'll fix):

```
msbuild path/to/MyExtension.csproj /t:PublishToMarketplace ...snip... /p:PublishExtension=path/to/the/extension.vsix
```

madskristensen:
Where should we document this? On the readme or on vsixcookbook.com?

reduckted:
> Where should we document this? On the readme or on vsixcookbook.com?

Good question. I was asking myself the same thing. I see that there's a "publishing" section on vsixcookbook.com. Perhaps we could add another sub-section related to build pipelines?

madskristensen:
It would be great to add it here perhaps: https://www.vsixcookbook.com/publish/marketplace.html

PR #270, opened 2022-01-18
Added code to limit the # of calls for DocumentEvents.Opened and Docu…

### LINKED ISSUE #269: Document events are called twice
The DocumentEvents.Closed  and DocumentEvents.Opened are called twice.
The implementation in the toolkit responds to the IVsRunningDocTableEvents.OnAfterFirstDocumentLock and IVsRunningDocTableEvents.OnBeforeLastDocumentUnlock events and according to the docs these events are called once for a ReadLock and once for an EditLock

Is it OK if I change the implementation and for example only call the events when the ReadLock changes ?

RobertvanderHulst:
…mentEvents.Closed to a single call per file. BeforeDocumentWindowShow is also only called for the first Show.

Finally added some extra documentation for the Opened and Closed events to explain that these events are also called for Project Files, and Solution Files and for the Miscellaneour Files project that is generated by VS.
Fixes #269


madskristensen:
Thank you so much!!

PR #271, opened 2022-01-20
Made some methods protected virtual so a subclass can returns its own…

RobertvanderHulst:
… completioncontext and can have its own IsValidBraceCompletionContext implementation


madskristensen:
Great catch

PR #277, opened 2022-01-22
Make properties settable by inheritors.

StevenRasmussen:
I am working on some changes to the `Community.VisualStudio.Toolkit.DependencyInjection` project that will allow registering commands in the DI container without the need to call the `BaseCommand.InitializeAsync` method.  I do however need to inherit from the `BaseCommand` and be able to set the `Package` and `Command` properties in the constructor.  This PR changes the setters on those public properties to `protected` so that inheritors can set them as well.


StevenRasmussen:
I also updated it so that the `BaseCommand<>` is now split between a generic and non-generic version.  This is a non-breaking change. Tt simply puts the static `InitializeAsync` and `InitializedCompletedAsync` methods on the generic version.  This allows inheritors to use the non-generic version and bypass the requirement to provide a constructor that takes 0 parameters. The dependency injection project commands will always have a constructor that takes 1 or more parameters.

madskristensen:
Thanks

Issue #279, opened 2022-01-24
Getting build errors with new Option page model


ErikEJ:
I am trying to use this in an exsting extension, but getting build errors - what am I doing wrong?

```
CreatePkgDef : error : ArgumentException: The page class EFCorePowerTools.AdvancedOptions must derive from DialogPage.
   at Microsoft.VisualStudio.Shell.ProvideOptionDialogPageAttribute..ctor(Type pageType, String pageNameResourceId)
   at Microsoft.VisualStudio.Shell.ProvideOptionPageAttribute..ctor(Type pageType, String categoryName, String pageName, Int16 categoryResourceID, Int16 pageNameResourceID, Boolean supportsAutomation, String[] keywords)
   at Microsoft.VisualStudio.Shell.ProvideOptionPageAttribute..ctor(Type pageType, String categoryName, String pageName, Int16 categoryResourceID, Int16 pageNameResourceID, Boolean supportsAutomation)
   at System.Reflection.CustomAttribute._CreateCaObject(RuntimeModule pModule, IRuntimeMethodInfo pCtor, Byte** ppBlob, Byte* pEndBlob, Int32* pcNamedArgs)
   at System.Reflection.CustomAttribute.CreateCaObject(RuntimeModule module, IRuntimeMethodInfo ctor, IntPtr& blob, IntPtr blobEnd, Int32& namedArgs)
   at System.Reflection.CustomAttribute.GetCustomAttributes(RuntimeModule decoratedModule, Int32 decoratedMetadataToken, Int32 pcaCount, RuntimeType attributeFilterType, Boolean mustBeInheritable, IList derivedAttributes, Boolean isDecoratedTargetSecurityTransparent)
   at System.Reflection.CustomAttribute.GetCustomAttributes(RuntimeType type, RuntimeType caType, Boolean inherit)
   at Microsoft.VisualStudio.Tools.CreatePkgDef.ProcessAssembly(String fileName, Hive hive, PkgDefContext context, Boolean register, RegistrationMode mode)
   at Microsoft.VisualStudio.Tools.CreatePkgDef.DoCreatePkgDef(InputArguments inputArguments)
   at Microsoft.VisualStudio.Tools.CreatePkgDef.Main(String[] arguments)	EFCorePowerTools			
```

```
internal class OptionsProvider
    {
        [ComVisible(true)]
        public class AdvancedOptions : BaseOptionPage<EFCorePowerTools.AdvancedOptions> { }
[...]

PR #284, opened 2022-01-26
Code analyzer to check the type passed to ProvideOptionDialogPageAttribute and ProvideProfileAttribute

reduckted:
Related to #279.

Using the wrong type in `ProvideOptionPageAttribute` seemed like a good candidate for a code analyzer to catch the problem early and to suggest a fix (where possible).

CVST003 will detect the wrong type being provided to `ProvideLanguageEditorOptionPageAttribute`, `ProvideOptionPageAttribute` and `ProvideToolboxPageAttribute` (they all expect a `DialogPage`)

CVST004 will detect the wrong type being provided to `ProvideProfileAttribute` (it expects an `IProfileManager`, which `DialogPage` implements).

Both diagnostics have a code fix that will suggest changing the type to a type that is nested within the type nested type that inherits/implements the correct type.  For example, give this:

```csharp
internal partial class OptionsProvider
{
    [ComVisible(true)]
    public class GeneralOptions : BaseOptionPage<General> { }

    [ComVisible(true)]
    public class OtherOptions : BaseOptionPage<Other> { }
}
 ```
and providing an options page like this:
```csharp
[ProvideOptionPage(typeof(OptionsProvider), nameof(TestExtension), "General", 0, 0, true)]
```

The code fix will suggest:
* Changing `typeof(OptionsProvider)` to `typeof(OptionsProvider.GeneralOptions)`
*  Changing `typeof(OptionsProvider)` to `typeof(OptionsProvider.OtherOptions)`

Demo:
code-fix

### DIFF
--- a/src/analyzers/Community.VisualStudio.Toolkit.Analyzers/Analyzers/CVST002DialogPageShouldBeComVisibleAnalyzer.cs
+++ b/src/analyzers/Community.VisualStudio.Toolkit.Analyzers/Analyzers/CVST002DialogPageShouldBeComVisibleAnalyzer.cs
@@ -31,8 +31,8 @@ namespace Community.VisualStudio.Toolkit.Analyzers
 
         private void OnCompilationStart(CompilationStartAnalysisContext context)
         {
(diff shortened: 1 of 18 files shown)


ErikEJ:
Sweet!

madskristensen:
That's awesome. Thank you

PR #300, opened 2022-02-10
Fixes an issue when trying to switch between projects for a shared file.

StevenRasmussen:
There has been an issue with regards to using the project dropdown when working with files from the shared project:
image

It appears that we are unable to `Import` the shared project using a `.props` file. Each project must reference the shared project individually.

You can see the issue reported to MS [here](https://developercommunity.visualstudio.com/t/systemargumentoutofrangeexception-specified-argume-2/1427702?from=email&viewtype=all#T-N1654857).


madskristensen:
This is so great. Thank you!

PR #309, opened 2022-03-15
Cancel project build if any project is canceled

jamesc-skyward:
If a project being built has the build canceled while building a project it is dependent on, the result never gets set. This pull request changes the code to cancel the result if any project is canceled. I think this is a safe assumption to make?

You can re-create this issue with the BuildActiveProjectAsyncCommand in the test extension. Build a project that is dependent on one or more projects and cancel the build before it gets to the active project. The build will be canceled, but you will not see any of the message boxes.


madskristensen:
Excellent. Thank you!

PR #314, opened 2022-03-23
Added IsDebuggingAsync() and GetDebugModeAsync() methods

jamesc-skyward:
There are times I don't want my extensions to be fired while the user is debugging. I added `VS.Debugger.IsDebuggingAsync()` to check whether or not the user is currently debugging. 

It can be used like this:
```
protected override void BeforeQueryStatus(EventArgs e)
{
    Command.Enabled = Package.JoinableTaskFactory.Run(async () =>
    {
        return await VS.Debugger.IsDebuggingAsync() == false;
    });
}
```
For good measure, I also added `VS.Debugger.GetDebugModeAsync()` for if you need more details.


madskristensen:
Great. thanks for the PR. I wonder if there's opportunity for simplification by calling this: `VsShellUtilities.GetDebugMode()`

madskristensen:
Awesome. Thanks

PR #316, opened 2022-04-06
Allow styles to be extended

reduckted:
Fixed #135.

You can now extend the custom Visual Studio-themed styles. This required a few changes:

First, the resource dictionary needed to be compiled like any other XAML file by including it as a `Page`.

Doing that meant the `xmlns` attributes for the `Microsoft.VisualStudio.Shell.15.0` assembly could no longer be used. This is because in the v14 version of the toolkit, that assembly doesn't exist, and is actually `Microsoft.VisualStudio.Shell.14.0`. We were previously working around that by loading the XAML file from the embedded resources, doing some string replacement, and then parsing the file as a XAML file. I've solved this problem be creating an internal class called `InternalResourceKeys` in which we effectively alias the properties from `Microsoft.VisualStudio.Shell.xx.0` that we need to use. 

For example, `VsResourceKeys.TextBoxStyleKey` defines the key of the style for the `TextBox` control that we extend in `ThemeResources.xaml` (previously accessed as `shell:VsResourceKeys.TextBoxStyleKey`). The `InternalResourceKeys` class has a property called `VsResourceKeys_TextBoxStyleKey` which simply returns the value of `VsResourceKeys.TextBoxStyleKey`, and in `ThemeResources.xaml` we now use `local:VsResourceKeys_TextBoxStyleKey`.

I've also added `CustomizedStylesDialog` to the demo extension to demonstrate how the styles can be extended. There are two examples:
* One extends the toolkit's `TextBox` style.
* The other extends Visual Studio's `Button` style.

You can access that dialog via the _Theme Window_ tool window.

### DIFF
--- a/demo/VSSDK.TestExtension/ToolWindows/ThemeWindow/CustomizedStylesDialog.xaml
+++ b/demo/VSSDK.TestExtension/ToolWindows/ThemeWindow/CustomizedStylesDialog.xaml
@@ -1,0 +1,61 @@
+﻿<platform:DialogWindow 
+    x:Class="TestExtension.CustomizedStylesDialog"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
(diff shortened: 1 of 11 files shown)


madskristensen:
Awesome. Thanks

PR #324, opened 2022-04-21
Add methods to remove attribute from Project

AchimStuy:
We are renaming some of our company's build properties. For that it would be nice to be able to remove properties.


madskristensen:
Nice one. Thanks!

PR #329, opened 2022-05-02
ShowError should only have an OK button and not a Cancel button

RobertvanderHulst:
As discussed on the forum


madskristensen:
Good call. Thanks!

PR #336, opened 2022-05-20
Add method for getting an output pane by its name

MariaSolOs:
This PR overloads `OutputWindowPane.GetAsync` to also allow getting an output pane by its name.

### DIFF
--- a/src/toolkit/Community.VisualStudio.Toolkit.Shared/Helpers/OutputWindowPane.cs
+++ b/src/toolkit/Community.VisualStudio.Toolkit.Shared/Helpers/OutputWindowPane.cs
@@ -3,7 +3,6 @@ using System.IO;
 using System.Threading.Tasks;
 using EnvDTE;
 using EnvDTE80;
-using Microsoft;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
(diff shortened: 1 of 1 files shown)


src/toolkit/Community.VisualStudio.Toolkit.Shared/Helpers/OutputWindowPane.cs
```diff
@@ -162,6 +165,30 @@ public static async Task<OutputWindowPane> CreateAsync(string name, bool lazyCre
...
+        /// Gets an existing Output window pane.
+        /// Returns null if a pane with the specified name does not exist.
+        /// </summary>
+        /// <param name="name">The pane's name.</param>
+        /// <returns>A new OutputWindowPane or null.</returns>
+        public static async Task<OutputWindowPane?> GetAsync(string name)
+        {
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
+            DTE2 dte = await VS.GetRequiredServiceAsync<DTE, DTE2>();
```
madskristensen, line 177: The toolkit must not use the DTE internally to avoid backwards compatibility with 32/64-bit versions

MariaSolOs: I see... I'll investigate if there's another way of getting a pane by its name using other interfaces.

MariaSolOs: So after reviewing the documentation and considering other options, I'm afraid that I don't know how to support this without using the `DTE`. That being said, I do see the value in this feature (and I'm also still a beginner in VS extensibility), and hence I opened [a new issue](#339) that can be appropriately labeled and may replace this PR.

madskristensen: Thanks Maria. I'm glad you've joined us here in the extensibility community and I wish you the best of luck and lots of fun as you continue extending Visual Studio. Never hesitate to reach out in case you need it!!

PR #340, opened 2022-05-25
Allowed a tool window or its content to get access to the ToolWindowPane

### LINKED ISSUE #335: ToolWindows: Ability to use Site services like IVsTrackSelectionEx
I migrated over from VSXtra to this toolkit, and I noticed during testing that my BaseToolWindow doesn't allow me to consume sited services like IVsTrackSelectionEx. The old ToolWindow implementation was based on WindowPane and thus allows me to use cited services.

My use case is that I have a tool window control and I want to set the property window selection to reflect a selected tool window item.

```cs
var trackSelection = Package.GetService<SVsTrackSelectionEx, IVsTrackSelectionEx>();

string browser = typeof(CustomPropertyBrowser).GUID.ToString("B");
trackSelection.OnElementValueChange((uint)VSConstants.VSSELELEMID.SEID_PropertyBrowserSID, 0, browser);
```

reduckted:
Fixes #335.

This change allows a tool window to access the underlying `ToolWindowPane` object, which can then be used to access sited services.

To get access to the pane, you must make your pane inherit from `ToolkitToolWindowPane` instead of `ToolWindowPane`.

If you need to access the pane in your `BaseToolWindow<T>` class, override the `SetPane` method. If you need to access the pane in your tool window's content, make your content (for example, your `UserControl`) implement the `IToolWindowPaneAware` class.

After the tool window is created _and_ the `ToolWindowPane` has been initialized by Visual Studio, the `BaseToolWindow<T>.SetPane` and `IToolWindowPaneAware.SetPane` methods will be called.

### DIFF
--- a/src/toolkit/Community.VisualStudio.Toolkit.Shared/Windows/ToolkitToolWindowPane.cs
+++ b/src/toolkit/Community.VisualStudio.Toolkit.Shared/Windows/ToolkitToolWindowPane.cs
@@ -1,0 +1,25 @@
+﻿using System;
+using Microsoft.VisualStudio.Shell;
+
+namespace Community.VisualStudio.Toolkit
+{
+    /// <summary>
+    /// An implementation of <see cref="ToolWindowPane"/> that allows the 
+    /// </summary>
+    public abstract class ToolkitToolWindowPane : ToolWindowPane
+    {
+        private bool _isInitialized;
+
+        /// <inheritdoc/>
+        protected override void Initialize()
+        {
+            base.Initialize();
+            _isInitialized = true;
+            Initialized?.Invoke(this, EventArgs.Empty);
+        }
+
+        internal bool IsInitialized => _isInitialized;
+
+        internal event EventHandler? Initialized;
+    }
+}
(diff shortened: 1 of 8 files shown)


madskristensen:
Awesome. Thanks. For this change, I should update the template pack to use the `ToolkitToolWindowPane` as well, right?

src/toolkit/Community.VisualStudio.Toolkit.Shared/Windows/ToolkitToolWindowPane.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using Microsoft.VisualStudio.Shell;
+
+namespace Community.VisualStudio.Toolkit
+{
+    /// <summary>
+    /// An implementation of <see cref="ToolWindowPane"/> that allows the
```
ejsmith, line 7: "Allows the" what?

reduckted: 😱 Oops! Thanks for spotting that.

reduckted:
> @madskristensen Awesome. Thanks. For this change, I should update the template pack to use the `ToolkitToolWindowPane` as well, right?

Yes, I think that would be a good idea. Tool windows will still work if the pane only implements `ToolWindowPane` though, so it's not super important to update it right away.

PR #342, opened 2022-05-31
Add cleaner command attribute check

MariaSolOs:
Refactors the `CommandAttribute` check in `BaseCommand` to use pattern matching and use `GetCustomAttribute` instead of `GetCustomAttributes`. This results in cleaner and safer code, since `attr` will only be defined when not null.


madskristensen:
Nice. Thanks

PR #357, opened 2022-06-25
Using InfoBar listener cookie

MariaSolOs:
When subscribing to UI events for an `InfoBar`, the code is currently discarding the listener cookie, but this cookie should be used to later unsubscribe the event sink when closing the gold bar.

I also made some nit changes on the XAML docs, keeping consistency across the repo (e.g.: [Here](https://github.com/VsixCommunity/Community.VisualStudio.Toolkit/blob/9c32305c83827d0bfd7c85777f9e79fc71e22c2b/src/toolkit/Community.VisualStudio.Toolkit.Shared/Commands/Commands.cs#L39) we use `<see langword="true" />` instead of `<c>true</c>`).


madskristensen:
Thank you!!

PR #365, opened 2022-07-27
Add virtual property TextMarkerTagType for SameWordHighLighterBase th…

RobertvanderHulst:
…at can be overwritten in a subclass to use a different TextMarkerdefinition, so the colors of the TextMarkers can be changed in a subclass.

Defaults to the old value of "MarkerFormatDefinition/HighlightWordFormatDefinition"

### DIFF
--- a/src/toolkit/Community.VisualStudio.Toolkit.Shared/MEF/SameWordHighlighterBase.cs
+++ b/src/toolkit/Community.VisualStudio.Toolkit.Shared/MEF/SameWordHighlighterBase.cs
@@ -20,6 +20,11 @@ namespace Community.VisualStudio.Toolkit
         [Import] internal ITextStructureNavigatorSelectorService? _textStructureNavigatorSelector = null;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/toolkit/Community.VisualStudio.Toolkit.Shared/MEF/SameWordHighlighterBase.cs
```diff
@@ -50,7 +55,7 @@ public class SameWordHighlighterBase : IViewTaggerProvider
 
     internal class HighlightWordTag : TextMarkerTag
     {
-        public HighlightWordTag() : base("MarkerFormatDefinition/HighlightWordFormatDefinition") { }
+        public HighlightWordTag(string tagName) : base(tagName) { }
```
madskristensen, line 58: Should we keep the default constructor to make it not break existing contracts?

RobertvanderHulst:
Mads,
This is an internal type. I was wondering why the constructor was public. That did not make sense to me.
I did not want to change that.
But I do not think there is a risk of breaking anything because of the visibility.

madskristensen:
Makes sense. Thanks!

PR #374, opened 2022-09-14
Updated XML comment for MessageBox.ShowConfirm() and suppressed a war…

RobertvanderHulst:
…ning for InfoBar.OnClosed


madskristensen:
Thanks

PR #384, opened 2022-10-17
Always try to write attribute to project file if property descriptor is not available

reduckted:
When setting a file attribute, if the `VSHPROPID_BrowseObject` property could be retrieved, but there wasn't a property descriptor for the specified attribute, then the attribute would not be set. Now, if will always fall back to writing the attribute to `IVsBuildPropertyStorage`.

Related to #383.


reduckted:
@madskristensen I think the build failed because it timed out downloading the .NET SDK. Are you able to trigger a rerun?

madskristensen:
Triggered a rerun

gmarginet:
Any idea when this will be released?

reduckted:
@madskristensen Any chance you could take a look at this?

madskristensen:
Thanks

PR #387, opened 2022-11-09
The HighLigth word feature stopped working after the window was split…

RobertvanderHulst:
… and unsplit or after the preview margin was shown because these all share the same buffer.

We now add a counter to the SameWordHighlighterTagger and when the Counter reaches zero then the tagger stops tagging. Also added a HighlightWord tagger to the example project that tags words inside text documents.


madskristensen:
Awesome. Thanks!

Issue #401, opened 2022-12-15
Recursively enumeration from Project gives me bin and obj dirs and its files


lsoft:
but these directories ARE NOT in the project, they are not included at all!

as a result I receive `.g.cs` files and have no ways to distinguish which `cs` files written by user, by source gen, or by other VS tooling.

I need a way to exclude bin and obj folder from enumerating. The difficult that these folders may have different names and different projects types may store such info in different ways....

Any ideas are very welcome!

```C#
        public static List<SolutionItem> ProcessDownRecursivelyFor(
            this SolutionItem item,
            SolutionItemType type
            )
        {
            var result = new List<SolutionItem>();

            if (item.Type == type)
            {
                var fi = new FileInfo(item.FullPath);
                //now I have something like this, but it's ugly: if (fi.Extension.NotIn(".dll", ".exe", ".obj", ".pdb", ".pkgdef", ".config", ".json", ".imagemanifest", ".vsixmanifest", ".vsix"))
                {
                    result.Add(item);
                }
            }

            foreach (var child in item.Children)
            {
                if (child != null)
                {
                    result.AddRange(child.ProcessDownRecursivelyFor(type));
                }
            }

            return result;
        }
```

lsoft:
Yes, I can to grab names for those folders:

```C#
            if(item is Community.VisualStudio.Toolkit.Project p)
            {
                var objFolder = await p.GetAttributeAsync("BaseIntermediateOutputPath");
                objFolder = objFolder?.Trim(' ', '\\', '/');

                var binFolder = await p.GetAttributeAsync("BaseOutputPath");
                binFolder = binFolder?.Trim(' ', '\\', '/');
```

so, probably, I can avoid enumerating inside them. But the question is still here: why the toolkit see these folders, if they are not in the project...

reduckted:
A `SolutionItem` is a wrapper around an `IVsHierarchy` - essentially it's an item that appears in Solution Explorer. If you have _Show All Files_ enabled, then the `bin/` and `obj/` folders appear in Solution Explorer and will therefore be included when enumerating.

As an aside, the `Properties` and `References` nodes are also included when enumerating, and are "virtual folders".

You can determine whether a file is only there because _Show All Files_ is enabled by getting the value of the `VSHPROPID_IsNonMemberItem` property.

[...]

PR #403, opened 2022-12-18
Adds the 'IsNonVisibleItem' property to the 'SolutionItem' class.

StevenRasmussen:
Addresses #401.  Adds a new lazy loaded property to the `SolutionItem` class called `IsNonVisibleItem`


src/toolkit/Community.VisualStudio.Toolkit.Shared/Solution/SolutionItem.cs
```diff
@@ -73,6 +83,11 @@ internal void Update(IVsHierarchyItem item)
         /// </summary>
         public IEnumerable<SolutionItem?> Children => _children ??= _item.Children.Select(t => FromHierarchyItem(t));
 
+        /// <summary>
+        /// Returns whether the item is normally visible in soltuion explorer
```
lsoft, line 87: @StevenRasmussen 1) typo `soltuion`; 2) also I guess the term `normally` can raise  questions for the programmer, may be add more context here? :)

lsoft (approved):

src/toolkit/Community.VisualStudio.Toolkit.Shared/Solution/SolutionItem.cs
reduckted: I would change "visible" to "hidden". Maybe something like this:

> Returns whether the item is normally hidden in solution explorer and only visible when Show All Files is enabled.

src/toolkit/Community.VisualStudio.Toolkit.Shared/Solution/SolutionItem.cs
```diff
@@ -38,9 +39,18 @@ internal void Update(IVsHierarchyItem item)
...
             ThreadHelper.ThrowIfNotOnUIThread();
             _item = item;
-            _hierarchy = item.HierarchyIdentity.IsNestedItem ? item.HierarchyIdentity.NestedHierarchy : item.HierarchyIdentity.Hierarchy;
-            _itemId = item.HierarchyIdentity.IsNestedItem ? item.HierarchyIdentity.NestedItemID : item.HierarchyIdentity.ItemID;
+            _hierarchy = item.GetHierarchy();
+            _itemId = item.GetItemId();
             _fullPath = GetFullPath();
+            _isNonVisibleItem = new Lazy<bool>(() =>
+            {
+                if (HierarchyUtilities.TryGetHierarchyProperty(_hierarchy, _itemId, (int)__VSHPROPID.VSHPROPID_IsNonMemberItem, out bool isNonMemberItem)
```
reduckted, line 47: This could be reduced to just :
```cs
return HierarchyUtilities.TryGetHierarchyProperty(_hierarchy, _itemId, (int)__VSHPROPID.VSHPROPID_IsNonMemberItem, out bool isNonMemberItem) && isNonMemberItem;
```

reduckted (changes requested):

src/toolkit/Community.VisualStudio.Toolkit.Shared/Solution/SolutionItem.cs
StevenRasmussen: Agreed. Thanks!

src/toolkit/Community.VisualStudio.Toolkit.Shared/Solution/SolutionItem.cs
StevenRasmussen: Definitely better. Thanks!

reduckted (approved):

madskristensen:
Awesome. Thanks!

PR #410, opened 2023-01-26
Allow command interceptors to be unregistered

reduckted:
This PR changes `Commands.InterceptAsync` to return a `Task<IDisposable>`. When you dispose of the returned object, the command interceptor will be unregistered.


madskristensen:
Awesome. Thanks!

PR #418, opened 2023-02-16
[Breaking Changes] Fix FrameShow enum

reduckted:
🚨 **This contains breaking changes** 🚨 

A few improvements to the `FrameShow` enum. Hopefully the breaking changes won't actually affect anyone. 🤞 

1. It was marked with the `[Flags]` attribute. The underlying `__FRAMESHOW` enum is not flagged.

2. There was a redundant `Unknown` member with a value of zero. That is equivalent to both `FRAMESHOW_Hidden` and `FRAMESHOW_WinHidden`. I've removed `Unknown`. **This is a breaking change.**

3. `FRAMESHOW_Hidden` is documented as being obsolete and has the same value as `FRAMESHOW_WinHidden`. I've removed `WinHidden` rather than `Hidden` because there is no corresponding `WinShown` value (but there is a `Shown` value). **This is a breaking change.**

4. I've added enum values for [`__FRAMESHOW2`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.__frameshow2?view=visualstudiosdk-2022), [`__FRAMESHOW3`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.__frameshow3?view=visualstudiosdk-2022) and [`__FRAMESHOW4`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.__frameshow4?view=visualstudiosdk-2022).

5. I've added a comment to the `<summary>` tag that specifies the underlying enum value. When I was trying to understand which value I should be using, I found myself searching the web for examples. First I had to discover the Visual Studio enum that `FrameShow` is actually wrapping, and then find the corresponding enum member name. It would have been much easier if it was all available in the documentation comments. 
 
ℹ️ One thing about the documentation comments that I noticed - when the enum is used within the same solution as the toolkit, intellisense actually shows the value assigned to the enum, like so:
image
But when you're referencing the actual NuGet packages, intellisense only tells you the numeric value:
image
Hence why I've added the underlying enum member to the documentation. 😄

### DIFF
--- a/src/toolkit/Community.VisualStudio.Toolkit.Shared/Windows/WindowFrame.cs
+++ b/src/toolkit/Community.VisualStudio.Toolkit.Shared/Windows/WindowFrame.cs
@@ -442,12 +442,8 @@ namespace Community.VisualStudio.Toolkit
     /// </summary>
     public enum FrameShow
     {
-        /// <summary>Reason unknown</summary>
-        Unknown = 0,
-        /// <summary>Obsolete; use WinHidden.</summary>
-        Hidden = __FRAMESHOW.FRAMESHOW_Hidden,
         /// <summary>Window (tabbed or otherwise) is hidden.</summary>
-        WinHidden = __FRAMESHOW.FRAMESHOW_WinHidden,
+        Hidden = __FRAMESHOW.FRAMESHOW_WinHidden,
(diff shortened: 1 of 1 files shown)


madskristensen:
Awesome. Thank you so much!

PR #420, opened 2023-02-19
Updated the nuget.org package source URL in the readme

reduckted:
See https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#package-source-sections

> Note that the source URL for nuget.org is https://api.nuget.org/v3/index.json.


madskristensen:
Thanks

PR #437, opened 2023-06-03
Get a WindowFrame wrapper for a tool window

reduckted:
This change allows a `ToolkitToolWindowPane` to get the underlying `WindowFrame` for the tool window. This is useful if you need to listen to events on the window frame to, for example, know when the tool window focused or closed.

### DIFF
--- a/demo/VSSDK.TestExtension/ToolWindows/RunnerWindow.cs
+++ b/demo/VSSDK.TestExtension/ToolWindows/RunnerWindow.cs
@@ -1,5 +1,6 @@
 using System;
 using System.ComponentModel.Design;
+using System.Diagnostics;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


yannduran:
Hi Dave,

This caught my eye because I'm looking for a way to "redraw" my toolwindow when a setting that affects the structure of the contents is saved. But unfortunately I don't see anything in `IVsWindowFrame` that will help with that.

Do you know of any way to force a toolwindow to be redrawn?

This is what I've tried:

In my toolwindow called `MainWindow`, I added this property:
```
   public class MainWindow : BaseToolWindow<MainWindow>
   {
       public static MainView View { get; private set; }

       // the rest of the code here
   }
```
Then I modified `CreateAync` to create an instance of `MainWindow` which I then save in the property before returning it: 

```
       public override Task<FrameworkElement> CreateAsync(int toolWindowId, CancellationToken cancellationToken)
       {
           View = new MainView();

           return Task.FromResult<FrameworkElement>(View);
       }
```

Later in the `Saved` method of a `BaseOptionModel` I have this code:

```
          try
          {
              MainWindow.View.InvalidateVisual();
          }
          catch (Exception ex)
          {
              _dialogService.ShowException(ex);
          }
```

But nothings happens :-( Any pointers you might have would be much appreciated. The solution may also be a good candidate to add to the toolkit.

It could be as simple as:

 try
          {
              MainWindow.Refresh(); // or .Redraw() etc
          }
          catch (Exception ex)
          {
              _dialogService.ShowException(ex);
          }

yannduran:
OK, thanks Dave. I'll look into that.

The datacontext is just a viewmodel class. But I think I might be trying to redraw the wrong object. I'm calling `InvalidateVisual` on the `MainWindow` class which inherits `BaseToolWindow<MainWindow>`. Perhaps I need to be calling it on the UserControl itself instead.

PR #438, opened 2023-06-06
Skipped the Newtonsoft.Json version when using the newer VS SDK.

### LINKED ISSUE #434: Visual Studio 17.6 uses Newtonsoft.Json 13.0.3
The `Newtonsoft.Json` version check [uses a maximum version of 13.0.1](https://github.com/VsixCommunity/Community.VisualStudio.Toolkit/blob/daf61228cb225b3f1c0220b50c250c9dcca92e5e/src/toolkit/nuget/build/imports/Constants.17.props#L4), but as of Visual Studio 17.6 (specifically [`Microsoft.VisualStudio.Sdk@17.6.36389`](https://www.nuget.org/packages/Microsoft.VisualStudio.SDK/17.6.36389)), version 13.0.3 is used.

reduckted:
Fixes #434 

Two changes:

1. If the `Microsoft.VisualStudio.SDK` package is greater than the most recent 17.5 version (which means it is 17.6.36389 or above), then the `Newtonsoft.Json` version check is skipped because NuGet will raise warnings if the version of `Newtonsoft.Json` is higher than it should be.
2. Setting the build property `CheckNewtonsoftJsonVersion` to true will completely skip the version check, just in case you don't want it to run for whatever reason.

### DIFF
--- a/src/toolkit/nuget/build/imports/NewtonsoftJsonVersionCheck.targets
+++ b/src/toolkit/nuget/build/imports/NewtonsoftJsonVersionCheck.targets
@@ -1,6 +1,6 @@
 <Project>
 
-  <Target Name="NewtonsoftJsonVersionCheck" AfterTargets="ResolveReferences">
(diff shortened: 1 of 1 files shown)


madskristensen:
Awesome. Thank you!!

PR #444, opened 2023-06-26
Make sure only a single SameWordHighlighter is attached to a textbuffer.

RobertvanderHulst:
Each TextView will register its events itself with the tagger from the textbuffer and when the textview is closed then the events are unregistered again.
The previous implementation was allocating way too many samewordhighlighters.
Also changed the color for the same word highlighters in the example code.


madskristensen:
Thanks

PR #508, opened 2024-08-15
QueryStatus is used to enable commands that are not enabled by defaul…

RobertvanderHulst:
…t, such as Goto Definition in the Sourcecode Editor


madskristensen:
Thanks

PR #524, opened 2024-10-30
Calculated the correct font size for use in WPF.

### LINKED ISSUE #523: FontSize inconsistency
The font size set and previewed in the VS Options dialog is not the same font size seen in the 'Fonts and Colors Windows'. To get the font as large as Environment font size 8, I need to set 'Top Left' font size to 12 (which looks huge in the preview in the Options dialog).

It doesn't seem to be DPI-dependent.

I believe some calculation is missing, but I have no idea what it might be (it's not px/pt)...
image

reduckted:
The font size specified in the `FontInfo` object is in points. WPF does not use points for font sizes, so the font size was always smaller than it should have been.

Fixes #523.


madskristensen:
Nice. Thanks

PR #554, opened 2025-05-01
Demonstrate how to treat a string property as an enum in Tools Options.

reduckted:
Demonstration for #553.


lsoft:
it work! 👍🫡

madskristensen:
Thanks

